Issue #15, opened 2015-12-23
Event Callbacks


markhayden:
### Objective
In order to make Pathfora more developer friendly when it comes to custom integrations we need to have more flexible callbacks for the various types of events the Pathfora widgets can trigger. This will allow a developer to handle journeys by creating cookies, passing data to alternate endpoints, etc.

### Scope
Implement ability to define callback function when Pathfora is initialized. Response should include at least two parameters: `type` and `payload` where type is a unique type for each of the types of event that can trigger the callback (e.g. click, loaded, form, etc.) and the payload contains data specific to that event as outlined below. This callback should be in addition to any native callback logic to provide events to lytics, google etc. 

### Tasks
* [x] Trigger Callback on Pathfora Load
* [x] Trigger Callback on Form Submission (form info / input values as payload)
* [x] Trigger Callback on Button Click (button info as payload)
* [x] Trigger Callback on Modal Open/Close (modal info / status as payload)

dylo0:
@markhayden for callbacks on Form submission, button clicks and modal close there's already existing mechanics for callback functions. From examples/actionWidgets.html:
``` javascript
    var modal = pathfora.Message({
        id: 'test',
        layout: "modal",
        msg: "Welcome to our website",
        confirmAction: {
            name: "Test confirm action",
            callback: function() {alert("test confirmation")}
        },
        cancelAction: {
            name: "Widget cancelled",
            callback: function() {alert("Cancel button clicked")}
        }
    });
```

Confirm/cancel actions are fired when user clicked ok / cancel button,  'name' is used for jstag tag. 

Below I'm pasting tracking widgets function so you can see which actions are reported under which names to jstag.send(). Basicly we push basic widget data like id type etc, and for all widgets, this + action names for confirmations/ cancels, and form data on contact forms.

```javascript
        trackWidgetAction: function(action, widget, htmlElement) {
            var params = {
                'pf-widget-id': widget.id,
                'pf-widget-type': widget.type,
                'pf-widget-layout': widget.layout,
                'pf-widget-variant': widget.variant
            };

            switch (action) {
                case 'show':
                    pathforaDataObject.displayedWidgets.push(params);
                    break;
                case 'close':
                    pathforaDataObject.closedWidgets.push(params);
                    break;
                case 'confirm':
                    params['pf-widget-action'] = widget.confirmAction.name;
                    pathforaDataObject.completedActions.push(params);
                    break;
                case 'cancel':
[...]

PR #20, opened 2015-12-29
Feature/15 event callbacks

mjurczyk:
### Issue #15 

*What's new:*

* Added 6 new event callbacks (as requested in the issue)
* Moved submit action logic from layout generator method to the action method
* Added docs for new callbacks

### DIFF
--- a/src/pathfora.js
+++ b/src/pathfora.js
@@ -1454,7 +1454,7 @@
             node: node
           });
         }
-        if (widget.config.layout === 'modal' && typeof config.onModalOpen === 'function') {
+        if (widget.config.layout === 'modal' && typeof widget.config.onModalOpen === 'function') {
           config.onModalOpen(callbackTypes.MODAL_OPEN, {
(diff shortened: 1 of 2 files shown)


src/pathfora.js
```diff
@@ -320,6 +331,16 @@
       } else if (condition.showOnInit) {
         context.pathfora.showWidget(widget);
       }
+
+//    NOTE theoretically onInit should be here, @see comment on initializeWidgetArray
```
mjurczyk: Discuss this.

mjurczyk:
Discuss if the onModalClose callback should be called when the form is submitted (whether we want to track non-effective closes or all of them).

araddon:
whoops, just commented on the ticket instead of here, but asked about global-level callbacks for any widget in the readme.

src/pathfora.js
araddon: thats fine

Issue #16, opened 2015-12-23
A/B Testing


markhayden:
# Objective
Simple a/b testing is a requirement for many marketing efforts. The goal with this enhancement is to allow the developer to select from two pre-defined testing methods natively in Pathfora: 100, 50/50.

# Scope
In order to add basic ability to a/b test pathfora widgets a few things must happen. First, we will need to accept an additional parameter when configuring Pathfora allowing the user to define which a/b testing method to use. 

By default we will assume no a/b testing (100), which leaves us with one initial other type (50/50). Keep in mind that down the road we might need to implement 33,33,33 or 20,80 but not at this point in time.

In addition to setting this param we will need a way to associate one or more Pathfora widgets to this a/b testing scope. Initially only two will be allowed (50/50) but this might increase in the future. Determining which widget to show can be done by choosing a random number between 1 and 2 or 0 and 1 and executing the widget by array key. 

Finally, we must ensure we store the selected test method for each device and load from there for all future versions so that the experience is consistent. 

(see event tracking ticket for further details on reporting)

# Tasks
* [x] Configure A/B Test Param
* [x] Associate Multiple Widgets by Test
* [x] Store Selected Test Version as Cookie
* [x] Prioritize Cookied Test Version if Exists

mjurczyk:
@markhayden please see https://github.com/lytics/pathforajs/blob/feature/16-ab-testing/AB_TESTING.md for more information about the feature, and please let me know if that's the implementation you requested.

Also, feature scales for future testing modes (33/33/33, 80/20 etc.), currently implements only 100 and 50/50.

Issue #23, opened 2016-01-04
Social Login (Facebook & Google)


markhayden:
### Objective
Social sign in simplifies the amount of data a user has to enter to fulfill forms, content gating, etc. We would like all widgets to support this type of one click data collection in addition to standard email / name fields as they exist today. For the initial launch we will focus on Facebook and Google while in a later version we will explore LinkedIn and possibly Twitter.

### Scope
**Facebook**
Add social sign in integration for Facebook. Button click should trigger an auth window in which we ask for `public_profile` and `email` as the scope of the auth. From this we should be able to get name, email and a handful of other public data that should be passed to Lytics / as a payload in the callback. 

https://developers.facebook.com/docs/facebook-login/web/login-button

**Google**
Google will function nearly exactly to the Facebook integration above, the only difference is the scope we will be seeking is `profile`. This agin should allow for basic contact information to be collected / passed as opposed to a full manual form for the end user. 

https://developers.google.com/identity/sign-in/web/sign-in

Examples of what these buttons will look like an how they might be used in execution can be found: https://github.com/lytics/pathforajs/issues/22

Once the inital support is in place and tested we can create additional tickets to roll the actual buttons out to all of the modal types, initially we will focus on the gate as referenced above.

### Tasks
* [x] Facebook social sign in integration.
* [x] Google social sign in integration.
* [x] Finalize data passed as payload (can review after we have a better idea of exact data we can get from the api)

PR #24, opened 2016-01-06
Feature/16-ab-testing

mjurczyk:
### Issue #16 

*What's new:*

* Added A/B testing feature (scalable for future modes, currently 100 and 50/50 implemented)
* Added docs for A/B testing usage
* Added A/B testing example file


AB_TESTING.md
```diff
@@ -0,0 +1,47 @@
...
+### `50/50`
+
+This mode divides users into two groups, each group with their separate set of widgets on the page.
+
+## Usage 
+
+In order to initialize A/B testing on the website, `setABTestingMode( mode )` has to be called before widgets are created on the page. For example:
+
+```js
+window.pathfora.setABTestingMode('50/50'); // Initialize a 50/50 AB testing
```
markhayden: Would it make sense / be possible to set the testing mode globally or per group? I would say its not super likely to start but we definitely could run into users running multiple widgets on a single page. If that occurs it looks like we can only have one a/b testing type?

This might be moot if running a `50/50` campaign on a group with only one widget would result in a `100` a/b test? since there is only one to show?

AB_TESTING.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+When you initialize a mode in which widgets are displayed to users according to their testing group, you should then specify a `testGroup` parameter on the widgets (if the parameter is not defined, widget will always be visible):
+
+```js
+var messageBar = pathfora.Message({
+  id: 'asd',
+  layout: 'bar',
+  msg: 'Welcome to our website',
+  testGroup: 0
```
markhayden: This is a little misleading in that it doesnt seem to group the tests together, rather its used to determine which widget gets triggered. It would be good to have a param to link specific widgets together as a group and another used as an identifier, this also becomes important when it comes to tracking. When a widget is selected that information can be passed to lytics as `groupname` `widgetid` or whatever so that we can report on what is working as well as what is not.

@araddon thoughts here?

markhayden:
I also may have left this out of the main scope doc but at any rate, it is important that the widget a/b testing information is passed along with the payload on form submit / button click etc. This way we can report on which widget of the a/b test is working and which is not. We'll likely need to incorporate a widget id of some sort, something that isnt different for each user and is a part of the widget config as noted above.

AB_TESTING.md
mjurczyk: Really good point, I will take care of the multiple testing instances on a single page right away :wink:

AB_TESTING.md
mjurczyk: @markhayden would you prefer it to be testing-oriented or widget-oriented then (both multiple groups on a single page and widget-to-test assignment) ? What I mean is:

Option 1 (widget-oriented approach, we assign tests to widgets same way we do now):
```js
var test1 = pathfora.ABTest('50/50');
var test2 = pathfora.ABTest('80/20');

var messageBar1 = pathfora.Message({
  id: 'a1',
  layout: 'bar',
  msg: 'test',
  testGroup: test1
});

var messageBar2 = pathfora.Message({
  id: 'a2',
  layout: 'bar',
  msg: 'test',
  testGroup: test2
});
```

Option 2 (test-oriented approach, we create tests from widgets):
```js

var messageBar1 = pathfora.Message({ id: 'a1' });
var messageBar2 = pathfora.Message({ id: 'a2' });

var test1 = pathfora.ABTest({
  type: '50/50',
  groups: [
    [ messageBar1 ],  // Group 1
    [ messageBar2 ]   // Group 2
  ]
});


var test2 = pathfora.ABTest({
  type: '80/20',
  groups: [
    [ messageBar1 ],  // Group 1
    [ messageBar2 ]   // Group 2
  ]
});
```

sarahyablok: A couple comments/questions:

- When the grouping formatting gets figured out, it would be nice to add to the documentation how to assign a test to a lytics segment. Also it would be nice if we could add this .md file to the wiki once this gets merged as that's where most of the documentation is.
- Is to have a test group assigned to the inverse option (https://github.com/lytics/pathforajs/wiki/Using-Lytics-segment-API) which gets triggered if the user doesn't belong to any of the target segments. More importantly if we could do a 50/50 test on that where half of the users get a widget, and the other half doesn't? Would be that possible?

dylo0: @sarahyablok good point with the wiki. For the questions I believe that Maciej will answer it tomorrow.
@mjurczyk we should update wiki pages for other already implemented features as well.

mjurczyk: @sarahyablok I will move docs to the wiki, I wasn't quite aware whether the wiki is used by anyone or up-to-date, so I just put the description in an md file.
Regarding the second question. I am not sure whether I understood correctly - If the problem is whether you can create a groups pair, in which first group has widgets and the second one has none - then yes, that would be possible.

markhayden: @mjurczyk I say we go with option 2, test based approach. Seems like this is perfectly aligned with how marketers would want to leverage a/b testing their strategies. Likewise it would give us a specific test name to leverage when we report data back to them. :+1:

mjurczyk: Thank you @markhayden , I will implement it that way then.

mjurczyk:
@markhayden this PR has been updated, please review. Sample usage, as in the example files:

```js
var ab1 = pathfora.ABTest({
  id: 'ab1',
  type: '50/50',
  groups: [
    [ modalA ]
    [ modalB ]
  ]
});

var ab2 = pathfora.ABTest({
  id: 'ab2',
  type: '80/20',
  groups: [
    [ messageBar ],
    [ /* Empty, no widget will be shown */ ]
  ]
});

window.pathfora.initializeABTesting([ ab1, ab2 ]);
window.pathfora.initializeWidgets([ messageBar, modal ]);
```

Would that be what we are aiming for?

src/pathfora.js
```diff
@@ -137,6 +137,60 @@
...
+      groupsNumber: groups.length
+    };
+  };
+
+  /*
+   * @global
+   * @property abHashMD5
+   * @description Hash used as the AB testing cookie (MD5 'ab')
+   */
+  var abHashMD5 = '187ef4436122d1cc2f40dc2b92f0eba0';
```
dylo0: @mjurczyk I'm wondering what do we gain by using this md5 hash here?

dylo0:
@mjurczyk I've done code review and it's fine. After @markhayden confirms that solution is what we're aiming we can merge that

src/pathfora.js
mjurczyk: Nothing in particular, just a cosmetic addition, instead of putting `test_1` etc. in the cookies.

markhayden:
It looks like if you use the pattern for initialize used in `segment widgets example.html` something doesn't get initialized quite right (see examples below):

```javascript
var messageBar = pathfora.Message({
  id: 'asd',
  layout: 'bar',
  msg: 'Welcome to our website',
  testGroup: 1
});

var modal = pathfora.Message({
  id: 'asd2',
  layout: 'modal',
  msg: 'Welcome to our website',
  confirmAction: {
    name: 'Test confirm action',
    callback: function() {alert('test confirmation')}
  },
  cancelAction: {
    name: 'Widget cancelled',
    callback: function() {alert('Cancel button clicked')}
  },
  testGroup: 1
});

var ab1 = pathfora.ABTest({
  id: 'ab1',
  type: '50/50',
  groups: [
  [ messageBar ],
  [ modal ]
  ]
});


var widgets = {
  target: [
    {
      segment: 'smt_new',
      widgets: [modal]
    },{
      segment: "smt_new",
      widgets: [messageBar]
    }
  ],
  inverse: []
};

var lyticsId = 1762;

config = {
  generic: {
    theme: "dark"
  }
};

window.pathfora.initializeABTesting([ ab1 ]);
```

for that base code if i initialize like: `window.pathfora.initializeWidgets([modal, messageBar], lyticsId, config);` it loads the widgets correctly but its not going to target based on the segment params defined.

if i do `window.pathfora.initializeWidgets(widgets, lyticsId, config);` it seems to only ever load the modal option. it does try to load the top bar but just results in a blank screen with no js errors.

we definitely need the ability to still define segments to target for each widget, they will not always be the same for an a/b test.

mjurczyk:
@markhayden I'll continue with that after I finish Social Plugins integration #23

dylo0:
@markhayden just to clarify: AB test groups should be 'above' Lytics segments - so we can specify one set of segment widgets to group A and the other set to group B. Correct?

markhayden:
did more testing on this one and as far as we can tell things are working. i think we are good to merge this version. we'll do some more thorough testing of advanced cases in the future but the initial use cases should be covered. thanks for you work on this. might be good to add some more detailed docs specific to a/b testing if possible, here is the config we ended up testing against.

```<script>
    var test1 = new pathfora.Message({
        msg: "Welcome to our site.",
        layout: 'modal'
    });

    var test2 = new pathfora.Message({
        msg: "Another headline option!",
        layout: 'modal'
    });

    var widgets = {
        target: [{
            segment: 'smt_unscored',
            widgets: [test1, test2]
        }]
    };

    var lyticsId = 1762;

    config = {
        generic: {
            theme: "dark"
        }
    };

    var ab1 = pathfora.ABTest({
        id: 'ab1',
        type: '50/50',
        groups: [
            [ test1 ],
            [ test2 ]
        ]
    });

    window.pathfora.initializeABTesting([ ab1 ]);
    window.pathfora.initializeWidgets(widgets, lyticsId, config);
</script>```

Issue #21, opened 2016-01-04
Event Tracking - Google Analytics


markhayden:
### Objective
Right now Lytics tracking is built in, which is great and useful. However, in the wild Lytics is likely to not be the only end point as it relates to tracking. Most of our customers use at a minimum one or two other forms of web tracking in addition to Lytics, such as Google Analytics. There needs to be some preliminary method to support this.

### Scope
For simplicity sake the first version will not require a config, we will simply check to see if the `ga` object exists, if it does we will assume they want events passed there and fire them off. Down the road this might be something controlled by a config, so keep that in mind.

On any action, everywhere jstag.send is fired, there will need to be an additional call for these integrations, Google Analytics initially.

That check will ensure the `ga` object exists, if it does it will fire an event per Googles event docs (https://developers.google.com/analytics/devguides/collection/analyticsjs/events).

```
ga('send', 'event', 'Lytics', 'Action Name', '', {nonInteraction: true});
```

The fifth param can just be an empty string for now as I don't think we have any details on the actions at this time.

Future integrations, not a part of this ticket but just food for thought will likely be Omniture, Marketo Munchkin, Mixpanel, KISS Metrics, Keen.io, Intercom, etc.

### Tasks
* [x] Ability to fire `ga('send')` event along side `jstag.send` events.
* [x] Integrated into all existing actions.

mjurczyk:
@markhayden PR submitted for this one, please review

dylo0:
@markhayden @mjurczyk - I'll do code reviews for this and A/B testing tomorrow, and merge it into develop branch. 
On Friday I'll ask our QA for checking develop branch and merge it into master after successful review

markhayden:
awesome. thanks @dylo0.

markhayden:
tested this one in development and all looked good. added a small pr to append the widget id to the label we send to GA just to facilitate better reporting but going to go ahead and close this.

PR #25, opened 2016-01-06
Feature/21 google analytics

mjurczyk:
### Issue #21 

*What's new:*

* Added Google Analytics integration

___

Google analytics is now automatically sent along with `jstag` calls. `ga` tracking function is called with the following parameters:

```js
'send',
'event',
'Lytics',
actionNameOrEventId,
'',
{
  nonInteraction: true
}
```

`actionNameOrEventId`parameter prioritises `name` param in the action, as in:

```js
var message = pathfora.Message({
  confirmAction: {
    name: 'action name',
    callback: function () {}
  }
});
``` 

For this example, fourth parameter will be equal `action name`. Otherwise, if name is not specified, `confirm` would be sent, as a default action id.

### DIFF
--- a/src/pathfora.js
+++ b/src/pathfora.js
@@ -1,4 +1,4 @@
-/* global jstag, pfCfg */
+/* global jstag, ga, pfCfg */
 "use strict";
 
 /**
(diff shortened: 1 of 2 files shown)


markhayden:
will test this today but looks solid. thanks for the quick turn.

src/pathfora.js
```diff
@@ -1244,6 +1245,21 @@
...
+          'event',
+          'Lytics',
+          data['pf-widget-action'] || data['pf-widget-event'],
+          '',
+          {
+            nonInteraction: true
+          }
+        );
+      } else {
+        // FUTURE Add exception
```
dylo0: @mjurczyk PR seems fine. You can merge it into develop after adding this exception

Issue #13, opened 2015-07-17
Exclude button from random layouts


dylo0:
Buttons are kind of another type, random widgets shouldn't pick it

mjurczyk:
@dylo0 Please review the PR, not sure if I understood the problem correctly.

PR #39, opened 2016-02-09
issue/13 Exclude button from random layouts

mjurczyk:
### Issue #13   

*What's new:*

* Removed `button` type from random layouts' options


dylo0:
:+1: after fixing conflitcs

PR #44, opened 2016-02-10
Issue/38 css improvements

### LINKED ISSUE #38: Design Cleanup
We need to go through the existing CSS and clean up the design a bit (the lytics team will pick up this ticket). Initial focus should be placed on `modal`, `slideout`, and `bar`, as those are the primary widgets in this first full integration. 

Need to ensure that everything looks good with/without image, headline, body, cancel, submit, etc. If necessary we can have two variations for each widget, 1 with image and 1 without. @bpatterson88  let me know when you are ready to kick this off and we can sit down to get everything running / run through stuff on your end.

bpatterson88:
closes #38


markhayden:
awesome. i'll look this over tonight / tomorrow morning.

src/less/widgets/widgets-general.less
```diff
@@ -199,8 +207,10 @@ a {
 .pf-subscription {
   form {
     overflow: hidden;
-    width: 100%;
-  
+    width: 80%;
```
sarahyablok: Ah. I can see why you did this, and then added 100% width to the slideout layout. I know we're not supporting folding just yet, but I think you'll find that that layout should also have 100% form width. It seems like it might make more sense to keep the general style 100%, and then just for the modal layout make it 80%? If that makes sense.

src/less/widgets/widgets-general.less
```diff
@@ -21,7 +21,7 @@ a {
 }
 
 .pf-box-shadow {
-  box-shadow: 0 0 40px rgba(0,0,0,.5);
+  box-shadow: 0 4px 4px -4px rgba(0,0,0,.5);
```
sarahyablok: I like the new shadow! Though I noticed on left or right aligned widgets it goes missing.
screen shot 2016-02-10 at 11 14 41 am
Probably not a big deal, but it would be nice to adjust the shadow for left or right positioned widgets. (This is in Themes example.html, for reference).

sarahyablok:
A couple notes on the bar layout. For a subscription widget the heights of the button and the input box I think could match.

screen shot 2016-02-10 at 11 22 17 am

Also for a plain message bar, it doesn't appear to be vertically center aligned.

screen shot 2016-02-10 at 11 24 40 am


Weirdly, it sets a `padding-top:7px` here, where on the subscription it's `padding-top:5px`, but I think both could have the `5px` padding. These are both artifacts from before so no fault to you!

markhayden:
appears to be something goofy going on with the positioning of the `bar` its under my top header instead of fixed to the top: 
image

Issue #7, opened 2015-05-26
Implement inline widgets


dylo0:
Future functionality, 
requirements:
1. Should be inserted into specified DOM element
2. Doesn't allow closing

PR #46, opened 2016-02-13
issue/7-inline-widgets: implemeneted new widget layout and position config

mjurczyk:
### Issue #7    

*What's new:*

* Added new inline `layout` for the widgets
* Added new `position` option for the widgets (DOM Selection)


dylo0:
:+1:

PR #50, opened 2016-03-11
Adds hideAfterAction functionality

sarahyablok:
We found that it would be useful to have the ability to permanently hide a widget after it was closed through a cancel, confirm, or close action. In each of the onclick functions, I've set a cookie `PathforaCancel_[id]`, `PathforaConfirm_[id]`, and `PathforaClose_[id]` respectively, to equal true. I also added a watcher function which will hide the widget if it detects said cookies. The config for this would look like:

```
displayConditions : {
  hideAfterAction: {
    cancel: true,
    closed: true,
    confirm: true
  }
}
```

This way we can choose to only show the modal until it is interacted with in some way. I think this could be improved later to be counts, and that way we can set something like:

```
displayConditions : {
  hideAfterAction: {
    closed: 5,
    confirm: 1
  }
}
```

Aka: the user can close the widget up to 5 times before it does not show up, or if they confirm once, never show it again.

The one I implemented that uses booleans just came about since we need it for a use case. Though if anyone wants to jump on implementing a more advanced version, feel free.

### DIFF
--- a/src/pathfora.js
+++ b/src/pathfora.js
@@ -522,13 +522,49 @@
     },
 
     registerHideAfterActionWatcher: function (hideAfterActionConstraints, widget) {
-      var valid = true;
-      var confirm = utils.readCookie('PathforaConfirm_' + widget.id);
-      var cancel = utils.readCookie('PathforaCancel_' + widget.id);
-      var closed = utils.readCookie('PathforaClosed_' + widget.id);
+      var valid = true,
+          now = Date.now(),
+          confirm = utils.readCookie('PathforaConfirm_' + widget.id),
+          cancel = utils.readCookie('PathforaCancel_' + widget.id),
+          closed = utils.readCookie('PathforaClosed_' + widget.id);
 
-      if ((confirm && hideAfterActionConstraints.confirm) || (cancel && hideAfterActionConstraints.cancel) || (closed && hideAfterActionConstraints.closed)) {
-        valid = false;
+      if (hideAfterActionConstraints.confirm && confirm) {
+        var parts = confirm.split(",");
+
+        if (parseInt(parts[0]) >= hideAfterActionConstraints.confirm.hideCount) {
+          valid = false;
+        }
+
+        if (parts[1] != undefined && (Math.abs(parts[1] - now) / 1000) < hideAfterActionConstraints.confirm.duration) {
+          console.log(Math.abs(parts[1] - now) / 1000 < hideAfterActionConstraints.confirm.duration);
+          console.log(Math.abs(parts[1] - now) / 1000);
+          console.log(hideAfterActionConstraints.confirm.duration);
+          valid = false;
+        }
+      }
+
+      if (hideAfterActionConstraints.cancel && cancel) {
+        var parts = cancel.split(",");
+
+        if (parseInt(parts[0]) >= hideAfterActionConstraints.cancel.hideCount) {
+          valid = false;
+        }
+
+        if (parts[1] != undefined && (Math.abs(parts[1] - now) / 1000) < hideAfterActionConstraints.cancel.duration) {
+          valid = false;
+        }
+      }
+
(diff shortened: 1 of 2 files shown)


test/pathforaSpec.js
```diff
@@ -1162,17 +1162,57 @@ describe('API', function () {
     var widget = $('#' + form.id);
     expect(widget.length).toBe(0);
   });
-  
-  if('should not show an impression counter widget without an id', function () {
+
+  it('should not show a widget with hideAfterAction without an id', function () {
+    expect(function() {
+      var form = new window.pathfora.Form({
```
markhayden: shouldn't need `window` on these i dont believe

test/pathforaSpec.js
```diff
@@ -1162,17 +1162,57 @@ describe('API', function () {
...
+      id: widgetId,
       msg: 'subscription',
       header: 'Header',
       layout: 'slideout',
       position: 'bottom-right',
       displayConditions: {
-        impressions: {
-          session: 1,
-          total: 5
+        hideAfterAction: {
```
markhayden: i almost wonder if we'll need to support a duration here as well (e.g. i want to hide the widget from anyone that clicked forever but those that have closed for only 7 days). maybe something like:

```
{
     "hideAfterAction": {
          "hide":true,
          "duration": 60 // minutes
     }
}
```

src/pathfora.js
```diff
@@ -915,7 +933,7 @@
       case 'inline':
         widgetCancel = widget.querySelector('.pf-widget-cancel');
         widgetClose = widget.querySelector('.pf-widget-close');
-        widgetOnModalClose = function (event) {
+        widgetOnModalClose = function (event, kind) {
```
markhayden: do we need kind here?

test/pathforaSpec.js
sarahyablok: Yeah, I had a similar note in the pr description, I think limiting the number of times it could show after each action too. So combining these into maybe something like:

```
hideAfterAction: {
  closed: {
    hideCount: 3,
    duration: 604800 // one week
  },
  confirm: {
    hideCount: 1
  }
}
``` 
In this case the modal would show again in a week after the first close, another week after the second, and then hide permanently (the length of the cookie) after the third close, however if the confirm button is ever clicked the modal will hide permanently (the length of the cookie).

Though now that I think about it I wonder if maybe this is too much, do you think people will care about the number of times they click "x"? Because the hideAfterAction duration would also work nicely with impressions. I'll implement what you suggested for now, but I'd like to get your thoughts on the above as well.

markhayden: i like the idea of more flexibility so i think the above is good. we'd just need a `*` or some sort of wildcard (maybe we just dont set a count) to show it indefinitely. 

so once a week until foreverness

src/pathfora.js
```diff
@@ -518,6 +521,57 @@
...
+
+      if (hideAfterActionConstraints.confirm && confirm) {
+        var parts = confirm.split(",");
+
+        if (parseInt(parts[0]) >= hideAfterActionConstraints.confirm.hideCount) {
+          valid = false;
+        }
+
+        if (parts[1] != undefined && (Math.abs(parts[1] - now) / 1000) < hideAfterActionConstraints.confirm.duration) {
+          console.log(Math.abs(parts[1] - now) / 1000 < hideAfterActionConstraints.confirm.duration);
```
markhayden: probably want to nix these console logs for production right?

sarahyablok: whoops! :ocean:

src/pathfora.js
```diff
@@ -924,8 +979,24 @@
           }
         };
 
+        updateActionCookie = function (name) {
+          var val = utils.readCookie(name),
+              duration = Date.now();
+          var ct;
```
markhayden: not a big deal but u can just `, ct;` here to maintain the var definition pattern

PR #51, opened 2016-03-16
Support for Automation & Bug Fix

markhayden:
Depends on https://github.com/lytics/pathforajs/pull/44 merging before.

Adds tooling and some helper functions to make the generation of a Pathfora config a bit easier on the Lytics product side. Adds a scaffold generator as well as insert helpers. This way we can better handle adding various widgets with various types of targeting (e.g. the same segment). Likewise it makes it a bit more managed on the Pathfora end as we start to push this out.

Also fixes small bug related to the `appearsOn` display property. It was functioning as an `and` rather than an `or` when multiple strings existed in the array. Updated that and added testing for all the above.


test/pathforaSpec.js
```diff
@@ -1083,7 +1083,7 @@ describe('API', function () {
     var widget = $('#' + form.id);
     expect(widget.length).toBe(1);
   });
-  
+
   it('should open site gating widget when the cookie is not set', function (done) {
     var gate = new window.pathfora.SiteGate({
       header: 'Blocking Widget',
```
sarahyablok: I think these should be `headline` now. Shouldn't cause any problems in the tests though.

src/pathfora.js
```diff
@@ -133,13 +133,26 @@
...
+      slideout: '<a class="pf-widget-close">&times;</a><div class="pf-widget-body"></div><div class="pf-widget-content"><h2 class="pf-widget-headline"></h2><p class="pf-widget-message"></p><a class="pf-widget-btn pf-widget-ok">Confirm</a><a class="pf-widget-btn pf-widget-cancel">Cancel</a></div>',
       bar: '<a class="pf-widget-body"></a><a class="pf-widget-close">&times;</a><div class="pf-bar-content"><p class="pf-widget-message"></p><a class="pf-widget-btn pf-widget-ok">Confirm</a><a class="pf-widget-btn pf-widget-cancel">Cancel</a></div>',
       button: '<p class="pf-widget-message pf-widget-ok"></p>',
       inline: '<div class="pf-widget-container"><div class="pf-va-middle"><div class="pf-widget-content"><a class="pf-widget-close">&times;</a><h2 class="pf-widget-header"></h2><div class="pf-widget-body"><div class="pf-va-middle"><p class="pf-widget-message"></p><a class="pf-widget-btn pf-widget-ok">Confirm</a><a class="pf-widget-btn pf-widget-cancel">Cancel</a></div></div></div></div></div>'
```
sarahyablok: Here too I think should be `<h2 class="pf-widget-headline"></h2>` Though i'm not sure it matters unless we support inline already.

sarahyablok:
:+1:

PR #53, opened 2016-03-23
Feature/44 exclude segs

markhayden:
Added ability to override lytics api and css file locally without modifying main pathfora.js file. Added testing build and default values for api. 

Added ability to exclude segments as well as testing for those scenarios and finally an example file for how to config the excludes. 

Desired behavior is to prevent displaying widgets for any users that have explicitly been excluded. for example, if i am in the "new_user" segment but also in the "converted" segment, we might want to suppress a message saying "make your first purchase". This could also be done through segments but that adds much more complexity.


src/pathfora.js
```diff
@@ -398,17 +399,29 @@
...
       }
+      if(!config.exclude){
+        throw new Error('Invalid scaffold. No exclude array.');
+      }
+
+      if (method === "target"){
+        subject = config.target;
+      }else if(method === "exclude"){
+        subject = config.exclude;
+      }else{
```
sarahyablok: Should we also check for `inverse` method here?

test/pathforaSpec.js
```diff
@@ -105,6 +105,59 @@ describe('Pathfora', function () {
     jasmine.Ajax.uninstall();
   });
 
+  it('bonk bnok bonok', function (done) {
```
sarahyablok: This test should have a real description.

src/pathfora.js
sarahyablok: We might also support the `common` method? :horse:

test/pathforaSpec.js
markhayden: hahahahahahaahahaha nice. i was struggling to get things to work, apparently i forgot to fix that once i got it straightened out :)

src/pathfora.js
markhayden: yah i wasnt even sure what common was doing so didnt include it in the scaffold.

sarahyablok:
:+1:

PR #54, opened 2016-03-24
Actually add class from className

sarahyablok:
A couple things I came across while working on the getty modal:

`className` didn't actually work (there was no way to add a custom class to a widget)

I also didn't like the fact that we were preemptively checking for a widget message. For example with getty we have an A/B test where the two modals are exactly the same except for the confirm button text. I couldn't set the message in the config object, which means I'd have to be setting it twice in both modal object, for example:

```js
var config = {
  message: {
    layout: "slideout",
    msg: "a message",
  }
}

var modal = new pathfora.Message({
  okMessage: "yay",
});

var modal = new pathfora.Message({
  okMessage: "blah",
});
```
that resulted in the error "Widget message is missing". I think we should allow msg to be set in the config, so I removed that check.

### DIFF
--- a/src/pathfora.js
+++ b/src/pathfora.js
@@ -1177,7 +1177,8 @@
         ' pf-widget-' + config.layout,
         config.position ? ' pf-position-' + config.position : '',
         ' pf-widget-variant-' + config.variant,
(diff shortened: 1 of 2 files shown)


markhayden:
:+1:

PR #61, opened 2016-04-01
Multiple display conditions

sarahyablok:
Change logic for display conditions and watchers to register all in an "AND" fashion rather than only allowing one.


markhayden:
:+1:

PR #62, opened 2016-04-06
Bug Fixes and CSS Updates

### LINKED ISSUE #56: Bugs/Features List
Here's a list of bugs to fix as we go through and document/test: (add more as necessary)

- [x] CSS: Form placeholder font for the message textarea is not getting set correctly.
- [x] Fix `top-fixed` and `top-absolute` positions for bar layout
- [x] Think about adding a pushDown (??) feature for top aligned bars, often they will cover a menu, if they float above it, maybe we could pass a element in to adjust the margin of?
- [x] Slideout should support bottom-left and bottom-right rather than left and right
- [x] add `origin` or something similar to slideout to determine which direction the slideout animates from.
- [x] Change `displayWhenElementVisible` to accept a string to query select instead of an id.
- [x] Small bug with impressions - where the count immediately gets incremented to 2. Does not affect getty

sarahyablok:
closes #56 

and all bugs/enhancements on that list.

Also a general styles clean up. I removed a bunch of unnecessary styles and pruned the following stylesheets:
- bars.less
- buttons.less
- modals.less
- slideouts.less
- widgets-general.less

CSS should now include a few annotations here and there as well. I removed all references to colors in the stylesheet as well, as far as I can tell they would always be overwritten with inline colors, either the default ones in pathfora, or in the user-defined config.


sarahyablok:
In retrospect I'm now learning that the css colors do matter, in one specific case: if the user defines the theme as "custom" without changing all or any of the fields, then it will resort to the default css colors. I think I'll add those back in, but they should match the default js theme.

src/less/widgets/bars.less
```diff
@@ -1,15 +1,19 @@
...
+
+  &.opened {
+    margin-top: 55px;
+  }
+}
+
 .pf-widget-bar {
-  background: @backgroundDefault;
-  color: @colorDefault;
   height: auto;
```
markhayden: we also need to adjust the z-index here to like 9999 or something super high. if tested against bootstrap its behind their default header styles (z-index 1030).

markhayden: bootstrap also does some wacky stuff with what i assume is the form styles:

sample

can probably address bootstrap stuff in another ticket / pr if it makes more sense.

markhayden:
this looks good :+1: outside of the z-index.

im gonna open a ticket for bootstrap and mobile testing so we can keep them separate.

PR #65, opened 2016-04-12
Update CSS, SiteGate, and Social Login

### LINKED ISSUE #63: Bootstrap Testing
Bootstrap seems to override a bunch of default styles that make the widgets break or appear oddly. The bar variety specifically:

sample

Lets do some thorough testing of all widgets on bootstrap as we will almost definitely run into many sites using bootstrap and pathfora at the same time.

sarahyablok:
closes #63 
- Tested The sdk with a bunch of bootstrap layouts (A few from here: http://getbootstrap.com/getting-started/#examples and a few fancy custom ones). Updated CSS to overwrite any common global styles that might interfere with the modal.

- Updates to SiteGate:
 - Traded first and last name for a single "Name" field
 - Fixed alignment of the input boxes
 - Added functionality for `fields` (see below)
 - Removed the "x" button so that it's truly a gate - can't be closed without submitting the form.
 - Added a cancel button which acts as a reset for the form.

- SocialLogin:
 - Should work now? Still pretty messy though, I don't like the way we're going about this. It might warrant rexamination later. Or @markhayden if you haveany ideas I can try and fix it up in this branch.
 - `showSocialLogin` function now works on siteGate and Forms. It will autofill if the person is already logged in, and show the button to login for google+ or facebook based on what's defined with `pathfora.integrateWith...`

- General Changes:
 - `required` now works with forms and siteGate. Format has changed to:
``` javascript
// Name and title fields are required, email and message are optional
required: {
  name: true,
  email: false,
  title: true,
  message: false
}
```
  - `fields` is the new `showName`, `showTitle`, etc:
``` javascript
// Only include the email and message field
fields: {
  name: false,
  email: true,
  title: false,
  message: true
}
```


dist/pathfora.js
```diff
@@ -948,26 +917,65 @@
...
-          // FIXME Cache
-          widgetForm.removeChild(widgetForm.querySelector('input[name="title"]'));
-        }
-        if (config.emailField === false) {
-          // FIXME Cache
-          widgetForm.removeChild(widgetForm.querySelector('input[name="email"]'));
+        if (config.showSocialLogin === false) {
+          node = widget.querySelector('.pf-social-login');
+
+          if (node.parentNode) {
```
markhayden: probably worth checking that node is not null here. if for some reason `.pf-social-login` doesnt exist this error.

dist/pathfora.js
```diff
@@ -1709,18 +1726,20 @@
...
 
-            window.gapi.signin2.render(signInBtn, {
+            window.gapi.signin2.render(signInBtn.id, {
               scope: 'profile',
-              width: 16,
-              height: 16
+              onsuccess: pathfora.onGoogleSignIn,
+              height: 25,
+              width: 90,
+              longtitle: false,
```
markhayden: extra comma at end

src/pathfora.js
```diff
@@ -948,26 +917,65 @@
...
+            return widget.querySelector('input[name="username"]');
+          return widget.querySelector('input[name="' + field + '"]');
         }
+
+        // Set placeholders
+        Object.keys(config.placeholders).forEach(function (inputField) {
+          var element = getFormElement(inputField);
+
+          if (element) {
+            element.placeholder = config.placeholders[inputField];
```
markhayden: i dont think this one will cause an error but is it safe to assume these are the right type to have placeholder or do we need to validate that its a field/input etc? might be over kill as long as it doesn't break on error.

src/pathfora.js
```diff
@@ -1709,18 +1726,20 @@
...
 
-            window.gapi.signin2.render(signInBtn, {
+            window.gapi.signin2.render(signInBtn.id, {
               scope: 'profile',
-              width: 16,
-              height: 16
+              onsuccess: pathfora.onGoogleSignIn,
+              height: 25,
+              width: 90,
+              longtitle: false,
```
markhayden: think i might have commented on the rendered version above but extra comma.

Issue #64, opened 2016-04-07
Mobile Testing


markhayden:
Need to figure out what / how we can support mobile browsers. 

As a step one lets just do a check and disable pathfora on mobile altogether. That will prevent anything from being published in a broken state.

* [x] shut pathfora off on mobile

Then we need to go in and test each widget and variation to make sure they render correctly on mobile / if we can support them all on mobile. 

* [ ] bar (top, bottom, scrolling)
* [ ] button (left, right, etc)
* [ ] slideout (bottom left/right)
* [ ] foldup (""  "")
* [ ] modal
* [ ] gate

also make sure we check each type (form, message, subscription)

PR #66, opened 2016-04-14
Bug fixes and Mobile Media Query

sarahyablok:
This addresses part 1 of #64 Added a simple media query to hide all widgets if the screen width is 480px or less. Definitely some of the larger widgets like modal still won't be looking good, so I'll address part 2 of that ticket soon. I just wanted to get this out before a potential deploy since I noticed a few bugs I didn't catch in the past PRs.


markhayden:
👍

PR #67, opened 2016-04-15
undefined fix

markhayden:
right now if for any reason there is no `msg` field it results in the string "undefined" getting published to the widget. we want to just omit the msg and never surface a js level error to markup.


sarahyablok:
👍

PR #68, opened 2016-04-18
add form example, field and required tests, fix bug

markhayden:
when removing the field we were not checking the boolean value so fields  with `true` were still getting removed. also added tests.


sarahyablok:
👍

PR #70, opened 2016-04-19
Sitegate / Form Fixes

markhayden:
sitegate was a little bit broken. it was closing on submit but data wasn't always getting passed to lytics, when it was it was using the placeholder object to get the inputs which was not passing name correctly. this led to more potential problems with our form processing in general.

fixed by iterating over children and determining if it can / should be sent from there. this will make it so in the future as we expand on forms we don't always have to explicitly define them in the pathfora library.

also made the inbound form data consistent (not using `sitegate` in name so we can establish some global lql to handle all pathfora form data.


sarahyablok:
👍

PR #72, opened 2016-04-21
Remove Comma from Cookie Values

sarahyablok:
Changed comma to an underscore for impression and hideAfterAction cookie values. This should comply with standard cookie value characters. We keep the support for reading cookies with a comma, but all new cookies are set and can be read with an underscore.


src/pathfora.js
```diff
@@ -611,8 +611,9 @@
       if (!total) {
         totalImpressions = 1;
       } else {
-        parts = total.split(","),
+        parts = total.split("_"),
         totalImpressions = parseInt(parts[0]) + 1;
+        parts = parts.length === 1 ? total.split(",") : parts;
```
markhayden: probably worth adding a comment here (and all places that do the same conversion) that notes the date we should end support for it. no reason to keep this around after all the pre-existing cookies expire.

src/pathfora.js
```diff
@@ -611,8 +611,9 @@
       if (!total) {
         totalImpressions = 1;
       } else {
-        parts = total.split(","),
+        parts = total.split("_"),
```
markhayden: total preference thing but you can also use `|` i believe. not a big deal since we control shit cookie but u are less likely to run into a pipe in key names and stuff like that in the grand scheme of cookie values. 

it also might make sense to just make a function to encode our cookies and decode them coming back out so they can have whatever values we want...again not a big deal for this fix but something to think about.

test/pathforaSpec.js
```diff
@@ -1276,7 +1276,7 @@ describe('API', function () {
 
   it('should not show widget if hideAfterAction duration not met', function () {
     var widgetId = 'hideAfterActionWidget1';
-    pathfora.utils.saveCookie('PathforaClosed_' + widgetId, "1," + Date.now());
+    pathfora.utils.saveCookie('PathforaClosed_' + widgetId, "1_" + Date.now());
```
markhayden: we should also add tests to test for the pre-existing cookie with comma case with comments on recommended removal date.

test/pathforaSpec.js
```diff
@@ -1369,6 +1369,39 @@ describe('API', function () {
     expect(widget.length).toBe(1);
   });
```
markhayden: in the spirit of being thorough it probably also makes sense to add a few tests to ensure that when a cookie is invalid it doesn't blow up pathfora / cause panics. 

when its missing an impression count or a value or is totally blank, etc. since cookies are public there is always a chance that someone will goof our cookie(s)

dist/pathfora.js
```diff
@@ -626,7 +628,7 @@
 
       if (valid && core.valid) {
         sessionStorage.setItem(id, sessionImpressions);
-        utils.saveCookie(id, Math.min(totalImpressions, 9998) + "," + now);
+        utils.saveCookie(id, Math.min(totalImpressions, 9998) + "|" + now);
```
markhayden: is 9998 here just an arbitrary max or am i derping?

markhayden: this isn't hurting n e thing, leaving for now but when we do a cleanup im pretty sure it can go.

test/pathforaSpec.js
markhayden: did this

sarahyablok:
👍

markhayden:
👍

PR #74, opened 2016-04-25
Add Expiration Date to Cookies

sarahyablok:
Most of the display condition etc, cookies weren't being set with an expiration date. I've now set any cookies to expire in a default of 365 days from when they are set.

I couldn't think of a way to write tests for this, since we can't check a cookie expiration date from `document.cookie` Any ideas @markhayden? I've gone through and tested the possibly scenarios manually, but if we could get an automated test that would be preferred.


src/pathfora.js
```diff
@@ -333,14 +333,12 @@
...
+    saveCookie: function (name, value, expiration) {
       var expires;
       var date;
 
-      if (days) {
-        date = new Date();
-        date.setDate(date.getDate() + days);
-        expires = '; expires=' + date.toGMTString();
+      if (expiration) {
+        expires = '; expires=' + expiration.toGMTString();
```
markhayden: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toGMTString

markhayden:
👍  after tests pass

PR #77, opened 2016-05-02
Add Support for Appears On Types

markhayden:
Adds support for:
* exact: uses full url, protocol and query params to evaluate
* simple: uses domain and path to evaluate
* string: looks for a string match against full url
* regex: evaluates regex agains full url

matching within the displays on property. also retains existing legacy array of strings approach.


sarahyablok:
👍

test/pathforaSpec.js
```diff
@@ -1583,6 +1583,220 @@ describe('API', function () {
...
+      id: 'e71c5416ac7345bcba8c5330d14c4a2e',
+      msg: 'subscription',
+      headline: 'Header',
+      layout: 'slideout',
+      position: 'bottom-right',
+      displayConditions: {
+        urlContains: [
+          {
+            match: 'exact',
+            value: 'http://localhost:9876/context.html'
```
sarahyablok: Idk if this will come up, but maybe make a comment somewhere here that if the test fails it could be due to karma not using the default port? :/

PR #79, opened 2016-05-02
remove initial passing of user data as seerid is not userid and is pa…

### LINKED ISSUE #78: Don't send the `seerid` as `user_id`
[Pathfora sends the `seerid` cookie value as the `user_id`](https://github.com/lytics/pathforajs/blob/cfaef69d04d4e8add2c71b62aefe8a927e90f975/src/pathfora.js#L1874-L1885). It would be nice if it didn't do that.

markhayden:
closes #78

initializeCustomAPI was passing the seerid as the user_id upon initializing the widgets. this was incorrect as the seerid is the uid and should never be merged with user_id. removed method entirely.


humanchimp:
👍

DingoEatingFuzz:
👍

PR #80, opened 2016-05-12
Alignment Fix

markhayden:
We were running into a small issue with alignment on the slideout modal if the body text was short:

http://mrtr.co/0B153V3X3X2n/Screen%20Shot%202016-05-12%20at%201.20.23%20PM.png
image

Added min-width to ensure that the div spans the width of the modal when copy isnt long enough to stretch it.


sarahyablok:
👍

PR #82, opened 2016-05-13
Prevent Border / Background on H2

markhayden:
We ran into some style conflicts on a site where the border and background image was set globally on H2s. This should prevent that as well as ensure alignment is as designed and not inherited.


sarahyablok:
👍

PR #86, opened 2016-05-20
v1 of Content Recommendations

### LINKED ISSUE #84: Content Recommendation Support (basic)
tbd

```
var promoWidget = new pathfora.Message({
  "layout": "modal",
  "variation": 3,
  "msg": "some cool content we know you'll be interested in",
  "recommend": {
    "query": {
      "raw": "FILTER AND (url LIKE 'www.getlytics.com/integrations/*') FROM content"
    },
    "content": [
      {
        "title": "My Page Title",
        "descr": "My page meta description",
        "img": "http://mypageimage.com/woot.jpg",
        "url": "http://google.com"
      }
    ]
  }
});
```

sarahyablok:
closes #84 

- Content recommendation support (variant 3) for `Message` modals with the `slideout` or `modal` layout. 
- Removes styles for images on modals with `variant: 1` this should enforce that variant 1 modals are text-only, variant 2 modals include images, and variant 3 include content recommendations.

View the [example file](https://github.com/lytics/pathforajs/blob/content-recommendation/examples/content-recommend.html) for a sample config and working example.

screen shot 2016-05-20 at 4 00 15 pm
screen shot 2016-05-20 at 4 00 39 pm


bpatterson88:
Design looks fresh 👍

araddon:
is there a widget layout that doesn't have the "check this out" title etc, instead is just 100% the content (which already has a title)?

sarahyablok:
@araddon Yes. Without a title or msg it looks like this:
screen shot 2016-05-20 at 6 30 11 pm

araddon:
nice! that looks great.

dist/pathfora.js
```diff
@@ -916,6 +913,39 @@
         switch (config.layout) {
         case 'modal':
         case 'slideout':
+          // Add Content recommendation if we have one
```
markhayden: might be better to do this as its own function. i can see future usecases for including content recommendation in a gated widget or a future `inline` type. having it broken out makes it easier to integrate into other places? 

also easier to write tests against probably

dist/pathfora.js
```diff
@@ -1568,7 +1604,7 @@
      * @throws {Error} error
      * @param  {array} array list of widgets to initialize
      */
-    initializeWidgetArray: function (array) {
+    initializeWidgetArray: function (array, accountId) {
```
markhayden: if we are going to force full account id after allowing aid we should do some validation here to surface a warning. should be able to just look for length and assume anything `<=` 4 is an aid that could have installed pathfora manually before the update.

dist/pathfora.js
```diff
@@ -2016,7 +2091,43 @@
...
+     * @description Fetch content to recommend
+     * @throws {Error} error
+     * @param {string} accountId  Lytics account ID
+     */
+    recommendContent: function (accountId, filter, callback) {
+      var seerId = utils.readCookie('seerid');
+      var recommendUrl;
+
+      if (!seerId) {
+        // set a default here, instead
```
markhayden: need this comment still? dont think we want pathfora changing the seerid since thats a function of jstag.

dist/pathfora.js
```diff
@@ -839,11 +834,13 @@
       var widgetBody = widget.querySelector('.pf-widget-body');
       var widgetMessage = widget.querySelector('.pf-widget-message');
       var widgetClose = widget.querySelector('.pf-widget-close');
+      var widgetContentUnit = widget.querySelector('.pf-content-unit')
```
markhayden: missing a `;` at the end

dist/pathfora.js
```diff
@@ -839,11 +834,13 @@
...
       var widgetTextArea;
       var widgetImage;
       var node;
       var i;
 
+
```
markhayden: extra space?

dist/pathfora.js
```diff
@@ -1601,10 +1637,49 @@
...
+            if (widget.content[0].default)
+              def = widget.content[0];
+            else
+              throw new Error('Cannot define recommended content unless it is a default');
+          }
+
+          api.recommendContent(accountId, widget.recommend.ql.raw, function(content){
+            var con = [];
+
+            if (content) {
```
markhayden: probably can be ignored for now but i wonder how long it will be before we need to store multiple pieces of content and do a random selection on the pathfora end of recommended?

dist/pathfora.js
```diff
@@ -1601,10 +1637,49 @@
...
+              con.push({
```
markhayden: you could probably just do an `unshift` to append the selected content to the array and avoid the need to re-add the default / overwrite the original content object? 

not sure if there is any value in retaining the default or not but worth mentioning.

src/pathfora.js
```diff
@@ -916,6 +913,39 @@
         switch (config.layout) {
         case 'modal':
         case 'slideout':
+          // Add Content recommendation if we have one
+          if (config.recommend && config.content) {
```
markhayden: should this fail or should we create the `config.content` array if it doesnt exist? if i dont set a default but i still have the recommend params we'd still want it to recommend and work right?

dist/pathfora.js
sarahyablok: There is also an option in the api `shuffle` which will randomize the order of the results. But yeah, we can think about how to do that stuff for v2. :)

dist/pathfora.js
sarahyablok: Yeah, I think the reason why I did that is because at some point prior to this, any array in the config gets turned into an object, so like if I console log `widget.content` before this it's gonna be `Object {0: Object}` no idea why it does that, but that's why I didn't do `unshift` however, it might make sense to just overwrite the default, as you were saying.

dist/pathfora.js
sarahyablok: Added!

dist/pathfora.js
```diff
@@ -1623,6 +1630,9 @@
         defaults = defaultProps[widget.type];
         globals = defaultProps.generic;
 
+        if (accountId && accountId.length <= 4)
```
markhayden: probably worth adding brackets here just to keep everything consistent?

markhayden:
👍  ship it

PR #88, opened 2016-05-24
Compile Templates w/ Gulp

markhayden:
Templates were included as a single string directly in pathfora.js. This presented a whole bunch of challenges when it comes to debugging, editing and writing / documenting custom CSS practices.

Pulls all templates out into a templates directory / their own .html file. Using gulp we bring in the files, build the templates object and insert it into the pathfora.js / .min.js files.


src/templates/form/folding.html
```diff
@@ -0,0 +1,24 @@
...
+		<p name="google-login" hidden></p>
+	</div>
+	<form>
+		<input name="username" type="text">
+		<input name="title" type="text">
+		<input name="email" type="email">
+		<textarea  name="message" rows="5"></textarea>
+		<button class="pf-widget-btn pf-widget-cancel">Cancel</button><button type="submit" class="pf-widget-btn pf-widget-ok">Send</button>
+	</form>
+</div>
```
sarahyablok: Probably not a big deal, but as far as I can tell everything else in pathfora is two-space indentation, so it might be nice to have the templates the same.

src/templates/subscription/bar.html
```diff
@@ -0,0 +1,9 @@
+<div class="pf-widget-body">
+</div>
+<a class="pf-widget-close">&times;</a>
+<div class="pf-bar-content">
+	<p class="pf-widget-message"></p>
+	<form>
+		<input name="email" type="email" required><input type="submit" class="pf-widget-btn pf-widget-ok" />
```
sarahyablok: This is not on you, and may be something for a separate pr, but I find it weird that some of the common templates have inconsistencies, like this is the only subscription template that uses a `input` vs. a `button` for the confirm button.

src/templates/subscription/folding.html
```diff
@@ -0,0 +1,19 @@
...
+<a class="pf-widget-caption-left">
+	<p class="pf-widget-headline"></p>
+	<span>&rsaquo;</span>
+</a>
+<div class="pf-widget-body">
+</div>
+<div class="pf-widget-content">
+	<p class="pf-widget-message"></p>
+	<form>
+		<button type="submit" class="pf-widget-btn pf-widget-ok">X</button>
```
sarahyablok: Also kind of weird that it defaults to "X" as the text for the submit button. I know it gets overwritten in the js, but ??? Again maybe there should be a separate pr where we uniform all the templates. I like how I'm noticing this all now that they're actually readable. 😆 👏

gulpfile.js
```diff
@@ -1,14 +1,26 @@
...
+    CSSURL;
+
+gulp.task('index', function () {
+  var target = gulp.src('./src/index.html');
+  // It's not necessary to read the files (will speed up things), we're only after their paths:
+  var sources = gulp.src(['./src/**/*.js', './src/**/*.css'], {read: false});
+
+  return target.pipe(inject(sources))
+    .pipe(gulp.dest('./src'));
+});
```
sarahyablok: Is this gulp task being used anywhere?

src/templates/subscription/bar.html
markhayden: totally agree, hopefully having these in a readable format will help us recognize some of these things. was thinking basically:

* get all the existing merges, minus this one merged
* add new templates to this pr and merge
* make a pass at existing templates to make them all consistent
* tackle mobile styles

more or less

src/templates/form/folding.html
markhayden: good call, i like consistency. ill change it.

gulpfile.js
markhayden: have to double check, doesnt look like it off hand but it was pre-existing.

sarahyablok:
👍

PR #93, opened 2016-05-28
Fix for Custom CSS - remove inline styles for default theme

### LINKED ISSUE #91: Remove the "default" theme in favor of CSS
When the user doesn't define a theme or custom colors in the config, we shouldn't be setting the default theme colors in js, we should rip these color definitions out of the js and put them in a pathfora base styles. This will make it much easier for a user to overwrite the base styles with custom CSS.

sarahyablok:
closes #91 

Plus a quick change to the cookie name for A/B Testing. (It's now `PathforaTest_[id]` instead of the hashed string prefix) We still recognize/support the old naming convention.


dist/pathfora.js
```diff
@@ -158,8 +147,8 @@
...
+    "modal": "<div class=\"pf-widget-container\"><div class=\"pf-va-middle\"><div class=\"pf-widget-content\"><a class=\"pf-widget-close\">&times;</a><h2 class=\"pf-widget-headline\"></h2><div class=\"pf-widget-body\"><div class=\"pf-va-middle\"><p class=\"pf-widget-message\"></p><div class=\"pf-social-login\"><p name=\"fb-login\" hidden></p><p name=\"google-login\" hidden></p></div><form><input name=\"username\" type=\"text\"> <input name=\"title\" type=\"text\"> <input name=\"email\" type=\"email\"><textarea name=\"message\" rows=\"5\"></textarea><button type=\"submit\" class=\"pf-widget-btn pf-widget-ok\">Send</button> <button class=\"pf-widget-btn pf-widget-cancel\">Cancel</button></form></div></div></div></div></div>",
```
markhayden: probably need to merge in develop and update these templates since they are now loaded from file.

src/pathfora.js
```diff
@@ -1434,53 +1401,75 @@
...
+          }
         }
       }
 
       if (contentUnitMeta) {
-        contentUnit.style.backgroundColor = colors.actionBackground;
-        contentUnitMeta.querySelector('h4').style.color = colors.actionText;
-        contentUnitMeta.querySelector('p').style.color = colors.text;
+        if (colors.actionBackground) {
+          contentUnit.style.backgroundColor = colors.actionBackground;
```
markhayden: can we be 100% sure that contentUnit is an actual element / has style and what not? if for some reason `.pf-content-unit` didnt exist this would blow up? ... even tho it should be there as long as our template doesn't get goofed.

markhayden:
one tiny thing then 🎉

PR #97, opened 2016-06-02
Fix CSS on default theme for recommendations

sarahyablok:
Noticed a small conflict between #93 and #86 where colors/styles for the content recommendation modal weren't getting added correctly for the default theme. This should fix.


markhayden:
👍

PR #99, opened 2016-06-08
Allow Unordered and Extra Query Params for String & Exact rule

### LINKED ISSUE #95: compare query params unordered
Query params are inherently unordered, so matching criteria should compare them in an unordered way.

`http://www.example.com?foo=1&bar=2` should match `http://www.example.com?bar=2&foo=1`.

Additionally, extra query params should be ignored.

`http://www.example.com?foo=1&bar=2&baz=3` should match `http://www.example.com?bar=2&foo=1`

sarahyablok:
closes #95 

Added functions to parse and compare query params for `urlContains` - includes the the default "legacy" and the `string`, and `exact` matching rule. For all comparisons we throw out the `lytics_variation_preview_id`.

### Examples
- User is visiting `http://localhost:9876/context?bar=2&foo=1&baz=3`
- exact match rule is `http://localhost:9876/context?baz=3&foo=1&bar=2`

**^ Match ^** 👍 

- User is visiting `http://localhost:9876/context?bar=2&foo=1`
- string match rule is `/context?foo=1&bar=2`

**^ Match ^** 👍 

- User is visiting `http://localhost:9876/context?bar=2&baz=3`
- string match rule is `/context?bar=2`

**^ Match ^** 👍 

- User is visiting `http://localhost:9876/context?bar=2&baz=3`
- exact match rule is `http://localhost:9876/context?baz=3`

**^ Not a Match ^** 👎


src/pathfora.js
```diff
@@ -505,14 +505,49 @@
       return true;
     },
 
-    pageVisitsChecker: function (pageVisitsRequired, widget) {
-      return (core.pageViews >= pageVisitsRequired);
+    /**
+     * @description Parse url queries as an object
+     * @param {string} url
+     */
+    parseQueries: function (url) {
```
markhayden: toootally nit picking here but this should probably be `parseQuery` since its only taking one.

src/pathfora.js
```diff
@@ -505,14 +505,49 @@
...
+     * @param {obj} matchQueries
+     */
+    compareQueries: function (queries, matchQueries) {
+      // any query in the matchQueries should exist in the url in the browser
+      for (var key in matchQueries) {
+        if (matchQueries[key] !== queries[key]) {
+          return false;
+        }
+      }
+
```
markhayden: we should also compare key counts here. this has potential to return a false truth if my `queries` has more keys than `matchQueries` but it contains all those present in the `matchQueries`

src/pathfora.js
```diff
@@ -505,14 +505,49 @@
...
     },
 
+    /**
+     * @description Compare query params between the url
+     *              the user is visiting and the match
+     *              rule provided
+     * @param {obj} queries
+     * @param {obj} matchQueries
+     */
+    compareQueries: function (queries, matchQueries) {
```
markhayden: also wondering (nitpicking) if we should be more explicit in naming these. its checking for an exact match and makes for a somewhat confusing read. 

maybe something like `qsAreEqual: function(subject, master)`

can probably come up with better naming but u get the idea.

src/pathfora.js
```diff
@@ -505,14 +505,49 @@
...
+        }
+      }
+
+      return true;
+    },
 
     urlChecker: function (phrases, widget) {
-      var url = window.location.href,
-          simpleurl = window.location.hostname + window.location.pathname,
+      var url = decodeURIComponent(window.location.href),
```
markhayden: whats the reason for `decodeURIComponent` here (or anywhere in the urls you are comparing)?

if my url contains encoded params it could cause things to break in a variety of ways. for example:
`https://www.getlytics.com/?bonk=%7B%22one%22%3A%22two%3F%22%7D`
is valid but decodes to
`https://www.getlytics.com/?bonk={"one":"two?"}`
which will break when you split on `?`

src/pathfora.js
```diff
@@ -505,14 +505,49 @@
...
+          simpleurl = decodeURIComponent(window.location.hostname + window.location.pathname),
```
markhayden: same thing here, dont think we want decodeURIComponent

src/pathfora.js
```diff
@@ -524,40 +559,52 @@
       // array of urlContains params is an or list, so if any are true evaluate valid to true
       if (phrases.indexOf('*') === -1) {
         phrases.forEach(function (phrase) {
+
           // legacy match allows for an array of strings, check if we are legacy or current object approach
           switch (typeof phrase) {
             case 'string':
-              if (url.indexOf(phrase) !== -1) {
-                valid = true;
+              var matchUrl = decodeURIComponent(phrase);
```
markhayden: same thing with `decodeURIComponent` here

src/pathfora.js
```diff
@@ -524,40 +559,52 @@
...
+
+              if (url.indexOf(matchUrl.split("?")[0]) !== -1) {
+                valid = core.compareQueries(queries, core.parseQueries(matchUrl)) && true;
```
markhayden: this one is doing a soft compare, whereas the other is looking for an exact match. might need to methods or a wrapper method or something to handle these better. maybe just a param to do the last check for length?

src/pathfora.js
```diff
@@ -524,40 +559,52 @@
...
               break;
+
             case 'object':
-              if(phrase.match && phrase.value){
+              if (phrase.match && phrase.value) {
                 switch (phrase.match) {
+                  // simple match
                   case 'simple':
-                    if (simpleurl === phrase.value) {
+                    if (simpleurl === decodeURIComponent(phrase.value)) {
```
markhayden: samesies

src/pathfora.js
```diff
@@ -524,40 +559,52 @@
...
                       valid = true;
                     }
                     break;
+
+                  // exact match
                   case 'exact':
-                    if (url === phrase.value) {
+                    if (url === decodeURIComponent(phrase.value)) {
```
markhayden: samesies

src/pathfora.js
```diff
@@ -524,40 +559,52 @@
...
                       valid = true;
                     }
                     break;
+
+                  // string match (default)
                   default:
-                    // default to string match
-                    if (url.indexOf(phrase.value) !== -1) {
-                      valid = true;
+                    var matchUrl = decodeURIComponent(phrase.value);
```
markhayden: samesies

src/pathfora.js
markhayden: after reading on i see why you didn't. some are comparing exact match and others aren't. probably need a param, wrapper or some way to evaluate the different cases.

src/pathfora.js
markhayden: we also need to take into account that we append a query param on preview. so basically need to toss out `lytics_variation_preview_id` in any of the comparisons or counts.

markhayden:
👍

PR #101, opened 2016-06-15
New Fields for SiteGate and Form

### LINKED ISSUE #90: Lead Capture Fields - WIP
As we gather feedback I'd like to make one or two passes through to add optional fields that are commonly requested. This is a placeholder ticket, don't start on it yet.

Add:
* First name
* Last name
* Email
* Message
* Company
* Job Title
* Phone Number

sarahyablok:
closes #90 

SiteGate and Form modules retain their default fields, though the following fields can now be added via the `fields` key for both module types: `phone`, `company`, `message`.

Also adds new doc page for `responsive` setting and added the grammar fixes from Jeff's review of the docs.


src/templates/sitegate/inline.html
```diff
@@ -10,11 +10,12 @@ <h2 class="pf-widget-headline"></h2>
...
           </div>
           <form>
-            <input class="pf-sitegate-field pf-field-full-width" name="username" type="text">
-            <input  class="pf-sitegate-field pf-field-full-width" name="email" type="email">
-            <input class="pf-sitegate-field pf-field-half-width" name="organization" type="text">
-            <input class="pf-sitegate-field pf-field-half-width" name="title" type="text">
-            <div class="pf-sitegate-clear"></div>
+            <input name="username" type="text">
+            <input name="email" type="email">
+            <input class="pf-field-half-width" name="company" type="text">
```
markhayden: one thing i was thinking about with these half widths is how do we handle the dynamic aspect? meaning we can technically shut off the title and leave the company on. in that case how does the form look? any way to make that better? maybe everything just needs to be full widget or something? not sure of the best way to handle. especially on the larger center modal.

sarahyablok: If you turn off one, the other goes full width. Don't worry I accounted for it. :)

markhayden: nm u already handled it (im reading backwards) !m

src/pathfora.js
```diff
@@ -85,13 +85,18 @@
...
+        phone: 'Phone Number'
       },
       required: {
         name: true,
         email: true
       },
-      fields: {},
+      fields: {
+        company: false,
+        phone: false,
```
markhayden: comma?

markhayden:
👍  we should update https://github.com/lytics/customers/blob/develop/public/web_default.lql#L56 with the new fields as well.

dist/pathfora.js
```diff
@@ -85,13 +85,18 @@
...
+        phone: 'Phone Number'
       },
       required: {
         name: true,
         email: true
       },
-      fields: {},
+      fields: {
+        company: false,
+        phone: false,
```
markhayden: comma

PR #104, opened 2016-06-17
escape URIs on output

humanchimp:
~~DO NOT MERGE YET~~

This PR aims to escape URIs on output using a method that will not re-escape already escaped parts of the text, making it safe to use without running the risk of double-escaping the text.

Some examples:

| Input | Output |
|-------------------|--------------|
|`https://www.getlytics.com/hello  world` |`https://www.getlytics.com/hello%20%20world`|
|`https://www.getlytics.com/hello %20world` |`https://www.getlytics.com/hello%20%20world`|
|`https://www.getlytics.com/hello%20 world`|`https://www.getlytics.com/hello%20%20world`|
|`https://www.getlytics.com/hello%20%20world`|`https://www.getlytics.com/hello%20%20world`|

Opening for discussion. I still need to get the tests passing/add some of my own.


markhayden:
👍 
in the description you have input `https://www.getlytics.com/hello world` and output `https://www.getlytics.com/hello%20%20world` but the output should be / is `https://www.getlytics.com/hello%20world` super nit but might be worth fixing just because

thanks for adding this

humanchimp:
there are actually 2 spaces there, but it's hard to see.

dist/pathfora.js
```diff
@@ -562,7 +626,7 @@
      */
     parseQuery: function (url) {
       var query = {};
-      var pieces = url.split('?');
+      var pieces = utils.escapeURI(url, { keepEscaped: true }).split('?');
```
sarahyablok: By calling this function here, I believe we'd only be escaping the queries. Thus:

- `http://localhost:8080/examples/_bar.html?hello=hi there`
would match
- `http://localhost:8080/examples/_bar.html?hello=hi%20there`

but: 
- `http://localhost:8080/examples/_bar hi.html`
would not match 
- `http://localhost:8080/examples/_bar%20hi.html`

I might be wrong, but it seems like the desired functionality is to escape the whole url.

humanchimp: ok, let me look into that

humanchimp: Yeah, you were totally right, of course. I updated the PR. I think I got it right this time.

sarahyablok:
👍

humanchimp:
I rebased this on develop and resolved the conflict about the trailing slash. Gonna push the button once green.

PR #105, opened 2016-06-17
Tests for A/B Tests!

### LINKED ISSUE #94: Add Tests for A/B Testing
Just realized we don't have any working tests for the A/B testing feature. 😱 

Lets add tests:

- [x] Only one group shows on an A/B Test
- [x] A/B test shows all modals in a group
- [x] Both groups work properly when the cookie is set appropriately
- [x] Multiple A/B Tests per page work
- [x] A/B Tests work with Audience targeting
- [x] Post - #93 merge - we support the old cookie naming convention
- [x] A/B Test `80/20` works
- [x] Any more error handling (?)
  - One widget should not be in more than one a/b test
  - id of ab-test should be unique

... feel free to add

sarahyablok:
closes #94 

for context, here is where the new, A/B Test tests begin: https://github.com/lytics/pathforajs/pull/105/files#diff-61d471d2018870c8111b90650fd28bc4R606

Also I took some time to ever so slightly organize the existing tests. They are a mess, so now they're at least in some logical order with comment descriptors above the categories. In the future we should obviously invest more time in organizing this.


markhayden:
👍

PR #108, opened 2016-06-22
Revamp of social Login

sarahyablok:
Complete revamp the way we do social Login buttons. Now both google and facebook buttons are completely custom styled to match the widget's appearance instead of using the default google and facebook sign in buttons.

screen shot 2016-06-22 at 12 34 43 pm

Also completely gutted the original code for these integrations. No more weird `setTimeout`s waiting for api variables to load, instead we use load callbacks supported by the google and facebook apis!


src/pathfora.js
```diff
@@ -1895,131 +1901,143 @@
...
+            core.onFacebookClick(fbBtns);
+          }
+        }
+      });
+    },
 
-            window.FB.XFBML.parse(widget);
-            signInBtn.className += ' social-login-btn';
+    /**
+     * @description Attempt to load forms' data from Facebook API.
```
markhayden: extra `'` in `forms'`?

src/pathfora.js
```diff
@@ -1895,131 +1901,143 @@
...
-                email: query.email || '',
-                firstName: query.first_name || '',
-                lastName: query.last_name || ''
-              });
-            });
-          }
-        });
+        if (typeof profile !== 'undefined') {
+          core.autoCompleteFormFields({
+            type: "google",
```
markhayden: total nit but looks like we use single quotes in most places but double here (linter will catch this stuff also if you are in the process of adding that)

markhayden:
couple super tiny nits but 👍 

we should also start to determine what browsers we want to support. i think with the pathfora sdk we have more freedom to cut off old browsers so we don't have to worry about things like svgs where the jstag library has to be much more forgiving on the data collection side.

sarahyablok:
@markhayden Agreed. Made a short ticket #112 to remind us of this.

PR #109, opened 2016-06-22
Exact Match should ignore trailing slashes

### LINKED ISSUE #107: Exact Match
When evaluating an exact match we need to check the last character of the url. If it is a `/` we need to try and evaluate it with and without the slash since the browser adds that in certain cases by default.

Also worth testing how this reacts when query params are present as well.

sarahyablok:
closes #107


markhayden:
👍

PR #111, opened 2016-06-23
Use Core Lytics Tag for Segments

markhayden:
As it stands pathfora was making its own unique call to the Lytics entity endpoint. This had two major downsides:

1) It put a dependency on the Pathfora Library. You had to have access to Lytics to really use it and if you didn't you would still be making an unnecessary api call with improper credentials.

2) Advancements made on the Lytics tag side to improve entity read times, first visit state processing and such were not shared with Pathfora. 

This leverages the existing infrastructure of Lytics Jstag to get the most accurate segments possible and removes the hard dependencies. 

@sarahyablok this will likely effect some of the working examples either here or in the docs so if you point me in the right direction i can get those configured as well.


docs/docs/targeting.md
```diff
@@ -8,15 +8,33 @@ var modules = {
   }]
 };
 
-pathfora.initializeWidgets(modules, 'YOUR LYTICS ACCOUNT ID');
+// using the lytics callback assumes that window.liosetup exists and the lytics js tag is loaded after the pathfora config
+window.liosetup.callback = function(){ pathfora.initializeWidgets(modules, 'YOUR LYTICS ACCOUNT ID'); };
```
sarahyablok: Can we just remove the lytics account id here and in all the examples on this page? Since it's not required for this feature anymore we wouldn't want people to think it's required here.

sarahyablok:
Docs wise we might also want to update this page. http://lytics.github.io/pathforadocs/api/acctid/ And/or move it elsewhere since now acct id is only necessary for content recommendations.

docs/docs/targeting.md
markhayden: done

markhayden:
good call and done.

docs/docs/targeting.md
```diff
@@ -8,15 +8,33 @@ var modules = {
...
 <img class="full" src="../assets/api_access.jpg" alt="Lytics Audience API Acess">
 
+## Leveraging the Lytics "Data Loaded" Callback
+
+When leveraging real-time user identity and/or segment membership in any Pathfora widget implementation, it is recommended that you use the Lytics "Data Loaded" callback. This ensures all user data has been loaded prior to the Pathfora initialization. Failing to load user identify may result in having users segments default to only the 'all' segment.
+
+``` html
+<script type="text/javascript">
+  window.liosetup = window["liosetup"] || {};
+  window.liosetup.callback = function(){ pathfora.initializeWidgets(modules, 'YOUR LYTICS ACCOUNT ID'); };
```
sarahyablok: forgot to remove the account id here.

sarahyablok:
👍

PR #114, opened 2016-06-24
JS Linting

sarahyablok:
Went through and cleaned all the js files using our new [eslint config file](https://github.com/lytics/pathforajs/blob/linting/.eslintrc). If you don't like any of the rules selected, speak now, because otherwise this is the new standard ❗ 

Also added a new gulp task `gulp lint` which will run the linter on all the necessary files.


src/pathfora.js
```diff
@@ -1437,24 +1412,25 @@
...
-          var rec = config.content[0]
+          var rec = config.content[0],
+              recImage = document.createElement('div'),
+              recMeta = document.createElement('div'),
+              recTitle = document.createElement('h4'),
+              recDesc = document.createElement('p');
+
           widgetContentUnit.href = rec.url;
 
-          var recImage = document.createElement('div');
```
markhayden: having the var definitions together at the top of the function makes for a difficult reading experience. probably worth adding comments where the definitions used to be to make it more readable.

src/pathfora.js
```diff
@@ -1283,21 +1258,21 @@
               ct;
 
           if (val) {
-            val = val.split("|");
+            val = val.split('|');
             // NOTE Retain support for cookies with comma - can remove on 5/2/2016
-            val = val.length === 1 ? val.split(",") : val;
-            ct = Math.min(parseInt(val[0]), 9998) + 1;
+            val = val.length === 1 ? val.split(',') : val;
+            ct = Math.min(parseInt(val[0], 10), 9998) + 1;
```
markhayden: this changed, intentional?

src/pathfora.js
```diff
@@ -1283,21 +1258,21 @@
               ct;
 
           if (val) {
-            val = val.split("|");
+            val = val.split('|');
             // NOTE Retain support for cookies with comma - can remove on 5/2/2016
```
markhayden: lets make another ticket to remove the cookie comma retention stuff but keep it separate / post the linting one.

src/pathfora.js
```diff
@@ -1165,29 +1150,19 @@
...
-      var widgetForm;
-      var widgetOnFormSubmit;
-      var widgetOnButtonClick;
-      var widgetOnModalClose;
-      var updateActionCookie;
-      var cancelShouldClose = true;
-      var i;
-      var j;
+      var widgetOk = widget.querySelector('.pf-widget-ok'),
+          widgetOnModalClose, updateActionCookie, widgetOnButtonClick;
```
markhayden: probably a nit but i wonder if we should make all variables with no value set consistently at the top of the var? some are above and some are below now.

src/pathfora.js
```diff
@@ -783,46 +770,47 @@
...
 
       if (hideAfterActionConstraints.closed && closed) {
-        var parts = closed.split("|");
+        parts = closed.split('|');
         // NOTE Retain support for cookies with comma - can remove on 5/2/2016
-        parts = parts.length === 1 ? closed.split(",") : parts;
+        parts = parts.length === 1 ? closed.split(',') : parts;
 
-        if (parseInt(parts[0]) >= hideAfterActionConstraints.closed.hideCount) {
+        if (parseInt(parts[0], 10) >= hideAfterActionConstraints.closed.hideCount) {
```
markhayden: intentional?

src/pathfora.js
```diff
@@ -783,46 +770,47 @@
...
 
       if (hideAfterActionConstraints.cancel && cancel) {
-        var parts = cancel.split("|");
+        parts = cancel.split('|');
         // NOTE Retain support for cookies with comma - can remove on 5/2/2016
-        parts = parts.length === 1 ? cancel.split(",") : parts;
+        parts = parts.length === 1 ? cancel.split(',') : parts;
 
-        if (parseInt(parts[0]) >= hideAfterActionConstraints.cancel.hideCount) {
+        if (parseInt(parts[0], 10) >= hideAfterActionConstraints.cancel.hideCount) {
```
markhayden: intentional?

src/pathfora.js
```diff
@@ -783,46 +770,47 @@
...
 
       if (hideAfterActionConstraints.confirm && confirm) {
-        var parts = confirm.split("|");
+        parts = confirm.split('|');
         // NOTE Retain support for cookies with comma - can remove on 5/2/2016
-        parts = parts.length === 1 ? confirm.split(",") : parts;
+        parts = parts.length === 1 ? confirm.split(',') : parts;
 
-        if (parseInt(parts[0]) >= hideAfterActionConstraints.confirm.hideCount) {
+        if (parseInt(parts[0], 10) >= hideAfterActionConstraints.confirm.hideCount) {
```
markhayden: assume these are all fine but just calling out to double check.

src/pathfora.js
```diff
@@ -138,7 +137,7 @@
 
   // NOTE HTML templates
   // FUTURE Move to separate files and concat
-  var templates = {{templates}};
+  var templates = '{{templates}}';
```
markhayden: pretty sure this will break the templates.

the linter will only work on the actual compiled file. anything uses the double brackets will throw an error. probably a way to ignore those in the config but this also skips the actual js that comes in for the templates which should follow the same patterns.

gulpfile.js
```diff
@@ -180,11 +181,18 @@ gulp.task('docs:mkdocs', shell.task([
...
+
+gulp.task('lint', function () {
+  return gulp.src(['src/pathfora.js', 'gulpfile.js', 'test/pathforaSpec.js', 'docs/docs/examples/**/*.js'])
+    .pipe(eslint())
+    .pipe(eslint.format())
+    .pipe(eslint.failAfterError());
+});
 
 gulp.task('test', ['build:styles', 'build:testjs']);
 gulp.task('build:local', ['build:styles', 'local:js']);
```
markhayden: probably worth adding the lint task to any of the build function since it should run against the compiled files not the source files. any dynamic aspects, like {{templates}} will not pass linter and wrapping in `''` will break the js once its rendered.

gulpfile.js
```diff
@@ -109,7 +110,7 @@ gulp.task('build:testjs', function () {
   gulp.src('src/*.js')
     .pipe(replace('{{apiurl}}', TESTAPIURL))
     .pipe(replace('{{cssurl}}', TESTCSSURL))
-    .pipe(replace('{{templates}}', prepareTemplates()))
+    .pipe(replace('\'{{templates}}\'', prepareTemplates()))
```
markhayden: this will work but it tells me we are linting the src file rather than the compiled file that includes templates and such. not sure we should be running it against the src files or changing the template defs. slightly worried that this would make it past tests accidentally since its going to be a string even if templates don't render. the other way will cause a panic.

sarahyablok: Yeah, I chose to lint the source, cause otherwise the line numbers are off on the linter output, which kind of makes it a pain to hunt down what actually needs to be changed. D:

gulpfile.js
sarahyablok: I actually originally had it in all of the build tasks, and it was a PAIN while developing later, so I decided to take it out. Maybe we need some task to run before you push up a commit? That way we're not always linting and it doesn't slow down the dev process but we ensure people actually lint before commiting? Idk, you should try it with the linting in the build task, and tell me if it's annoying to you.

markhayden: yah it might be annoying in the build task that watches for sure. we/i can just make a new task to run before we merge anything. or maybe we include it with tests only?

gulpfile.js
markhayden: true :). for jstag im just doing the src till i get most of it fixed then switching to the dist.

src/pathfora.js
DingoEatingFuzz: > If radix is undefined or 0 (or absent), JavaScript assumes the following:

>  - If the input string begins with "0x" or "0X", radix is 16 (hexadecimal) and the remainder of the string is parsed.
>  - If the input string begins with "0", radix is eight (octal) or 10 (decimal).  Exactly which radix is chosen is implementation-dependent.  ECMAScript 5 specifies that 10 (decimal) is used, but not all browsers support this yet.  For this reason always specify a radix when using parseInt.
>  - If the input string begins with any other value, the radix is 10 (decimal).

From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

gulpfile.js
sarahyablok: Looks like there is a way to disable rules for a single line http://eslint.org/docs/user-guide/configuring.html#disabling-rules-with-inline-comments. Since this templates line is the only one causing problems in `src` we could disable the rule for this line, and still lint the src code? Thoughts?

sarahyablok: Ignore that last comment. Apparently you can only ignore lines for rules, and not parsing errors. D:

src/pathfora.js
sarahyablok: Added a few, though for the most part a lot of the vars were already defined at the top even before the linting. Let me know if/where you see places that may need more context.

src/pathfora.js
sarahyablok: Added a ticket #117

gulpfile.js
sarahyablok: Ok, for now I've decided to re-add it to the build task, but it doesn't run on the watcher. That way it'll run with tests. If we find a better solution later we can change it.

markhayden:
👍

PR #116, opened 2016-06-28
Migrate Tag Level Personalization

### LINKED ISSUE #115: Migrate Tag Level Personalization from Lytics Tag
Details here: https://github.com/lytics/lio/issues/7362

sarahyablok:
closes #115 
Adds new class `Inline` to handle all the inline personalization elements, plucked and slightly modified the existing code from the jstag. This should accompany a pr in lio modifying the jstag.

@markhayden We should think about how we want to document this, we could move some of the examples from https://github.com/lytics/examples in here. Though I'd be hesitant to move the squeaky fuzzy site in since it's got a bunch of extra demo tools in it, which is great, but maybe a bit heavy for these developer docs. Perhaps we should think about keeping the docs/examples here basic, and dev focused while building out a better demo tool for sales/marketing that is more squeaky fuzzy-esque. If you would like me to do a quick write up of docs for this feature I can add it to this pr.


dist/pathfora.js
```diff
@@ -6,7 +6,7 @@
  */
 (function (context, document) {
   // NOTE Output & processing variables
-  var Pathfora, utils, core, api;
+  var Pathfora, utils, core, api, Inline;
```
markhayden: cap?

dist/pathfora.js
```diff
@@ -2772,9 +2778,131 @@
     this.utils = utils;
   };
 
+  /**
+   * @class
+   * @name Inline
+   * @description Inline Personalization
+   */
+  Inline = function () {
+    this.lioElements = [];
```
markhayden: probably dont need namespacing here and if we do it should e related to pathfora and not lio since the two libraries are technically independent.

markhayden:
not sure if this is the best place for this or not but:

Lets talk callbacks.

Historically we have only had support for one callback when the entity is loaded from the `/me` endpoint. That was added using the window.liosetup object as described below:

``` javascript
window.liosetup = window.liosetup || {};
window.liosetup.callback = function; // single function or array
```

Recently we introduced a small change in lio.js that accepts that callback as either a single function or an array of functions but historically it has always been a single function, so things will be in the wild as singles.

Now with Pathfora we have a need to add additional callbacks on top of whatever the user may have manually defined on their own. The other variable we need to account for is its technically possible for pathfora to be initialized before lio, therefore any calls to `lio` itself will fail. 

Possible solution?:
* Add an array to hold callbacks within Pathfora if `window.lio` is undefined.
* Add an array to hold callbacks in Lio.
* Upon Lio being initialized have it pull all callbacks from window.liosetup as well as pathfora, if it exists, into the master callback array.
* On first load it handles all those callbacks and removes them from the array.
* Every call to lio after that should be able to use `window.lio.data` to access the entity data removing need for additional callbacks.

Future things to ponder:
* Load time is/will be an issue. Need to figure out how we want to handle fallbacks and if there is ever any need to give each callback a specific wait limit or something like that. If I am doing modals with pathfora a second wait may not matter. If I am doing inline personalization either custom or with Pathfora that delay becomes more important and we may need to fallback to cookie after 500ms or something to that effect.

sarahyablok:
@markhayden Updated both this pr and the one in lio. So they should be ready to review again. I also changed the callback for initialize instead of forcing the user to set the callback manually like so,
``` js
window.liosetup.callback = function () {
  pathfora.initializeWidgets(modules);
};
```

`initializeWidgets` will now add a callback for you. https://github.com/lytics/pathforajs/blob/3de591d338433bc3266140faa78e08bf48dd0c0c/dist/pathfora.js#L2377 And execute the meat of the initialization only once the segments are loaded by lio. The only thing to note is if they don't load the jstag in this case the widgets will never be initialized. Which I think is the desired behavior anyway? But I made note of it in the docs.

dist/pathfora.js
```diff
@@ -2302,6 +2308,28 @@
...
+     */
+    this.callbacks = [];
+
+    /**
+     * @public
+     * @description Add callbacks to execute once segments load
+     */
+    this.addCallback = function (cb) {
+      if (context.lio) {
+        if (!context.lio.segmentsCbCalled) {
```
markhayden: instead of adding another new variable to lio we can just use `lio.loaded` it gets set to true after the entity as been loaded. at that point we can just just fire the callback `cb(window.lio.data)`. 

basically if lio has been loaded we dont want to add any more callbacks since there isnt an event to fire them / the data is already accessible.

dist/pathfora.js
```diff
@@ -2344,49 +2372,55 @@
...
-            // handle inclusions
-            if (widgets.target) {
-              tl = widgets.target.length;
-              for (ti = 0; ti < tl; ti++) {
-                target = widgets.target[ti];
-                if (segments && segments.indexOf(target.segment) !== -1) {
-                  targetedwidgets = target.widgets;
+
+          // Add callback to initialize once we know segments are loaded
+          var cb = function () {
```
markhayden: do we need to pass the data object as an argument here?

dist/pathfora.js
sarahyablok: Oh sweet, I didn't know there was a variable for this already. I will fix in both PRs. :)

dist/pathfora.js
```diff
@@ -2292,6 +2291,118 @@
...
+        this.elements = document.querySelectorAll('[' + attr + ']');
+      }
+
+      var elements = this.elements;
+
+      for (var i = 0; i < elements.length; i++) {
+        if (elements[i].getAttribute(attr) !== null) {
+          var theElement = elements[i],
+              group = theElement.getAttribute('data-liogroup'),
+              type = theElement.getAttribute('data-liotype'),
```
markhayden: totally minor and your call but would it be better to rename all these attributes now that they are outside of lio? can probably do something a bit more clean like:

`pfa-group`
`pfa-type`
etc?

src/pathfora.js
```diff
@@ -2256,6 +2255,118 @@
...
+      }
+
+      return dataElements;
+    };
+
+    /*
+     * @description Show/hide the elements based on membership
+     */
+    this.procElements = function () {
+      var elementObj = this.prepElements('data-liotrigger');
```
markhayden: wondering if we should skip everything that follows if `elementObj === {}` so we dont alter the DOM when unnecessary. Seems like we would do a ton of processing / adding CSS even if there were no instances of `data-liotrigger`.

sarahyablok: We could do this, but currently the CSS addition is in the constructor for inline should actually get added before `inline.procElements` is ever called.  I actually think this is a good thing though. `inline.procElements` is called only once the dom content has been loaded https://github.com/lytics/pathforajs/blob/migrate-outlio/src/pathfora.js#L2416 and if we tried to look for the existence of `data-liotrigger` elements before adding the css to hide them, the dom content would have to already be loaded, thus I think we'd get that annoying flicker on load, where the element is shown and then quickly hidden by the css addition. :/

dist/pathfora.js
sarahyablok: Will do, I will update this branch with the name changes and then make a ticket to change them in `www` once this gets deployed.

src/pathfora.js
markhayden: ah good point. carry on.

markhayden:
👍

PR #120, opened 2016-07-06
Inline Content Recommendations v1

### LINKED ISSUE #119: Inline Content Recommendations v1
With the migration of the datalio inline element toggle from the jstag to pathfora, and since the topic of inline recommendations have come up quite recently, I thought that it might be useful to implement content recommendations in a similar style for us to use on our marketing site, and as a potential early option for customers with dev resources. Here's the idea for a template:

``` html
<div data-lioblock="block1" data-liorecommend="www.getlytics.com/blog/posts/*">
  <div data-liotype="image"></div>
  <a data-liotype="url"><h2 data-liotype="title"></h2></a>
  <p data-liotype="description"></p>
</div>
```

Each recommendation block has a unique `data-lioblock` name. `The data-liorecommend` attribute provides a string url pattern to be used as a basic filter for the recommendation. In the future we could implement this filter in a better way, but to start we would only support url filtering. Multiple recommendation blocks with the same `data-liorecommend` attribute will be filled with results from a single same api response, so as not to repeat content. `data-liotype` dictates the content that should be filled in via javascript.

- `image` = primary image of the document. (Set as background image if on a `div` or the `src` attribute if on an image.
- `url` = url of the document. (Set as `href` attribute if on a link, or innerHTML if on a div).
- `title` = title of the document.
- `description` = meta description of the document.

We will also support a default option should the recommendation request fail.

Here's an example (with minimal styles applied) to show what these recommendations make look like.

screen shot 2016-07-05 at 12 35 07 pm

sarahyablok:
closes #119

screen shot 2016-07-05 at 12 35 07 pm

We should discuss the following before reviewing:

- [x] A better way to get the acctid from jstag? Right now I'm setting it with `window.ly_cid` which seems gross, we should probably have a better way to access this.

- [x] ~~Need to check with data science, but it seems like filters are broken on the recommendation API at the moment? Either that or I am inputing it wrong. the image above was the results from no filter, but when I add a basic URL filter it won't return any results, and thus goes to default.~~ This isn't a problem anymore as far as I can tell!

- [x] We should merge #116 before this since everything in there is in here too. This PR builds off of that code.


markhayden:
pr on /tag endpoint is in review but it will make the account id available at `lio.account.id` 

should be much more reliable than the ly_cid method as it will always be full account id delivered by our api rather than user input.

sarahyablok:
@markhayden Updated the pr to use `lio.account.id` It's also important to note that I've used that for the original content recommendation pathfora modals now as well. Therefore we've removed the Lytics Account ID as a second parameter to the `initializeWidgets` function and should update any dependencies accordingly.

src/pathfora.js
```diff
@@ -2426,10 +2583,9 @@
      * @public
      * @description Initialize Pathfora widgets from a container
      * @param {object|array}   widgets
-     * @param {string}         lyticsId
      * @param {object}         config
      */
-    this.initializeWidgets = function (widgets, lyticsId, config) {
+    this.initializeWidgets = function (widgets, config) {
```
markhayden: this is totally fine / much better but we should try and make it as backwards compatible as possible. should be able to just check if that second argument is actually a config and not a string, if its a string we should assume its an id and just throw it out and look to the 3rd argument for the config.

definitely should update all of our internal stuff to correctly handle this but difficult to ensure everyone that could be using this will also update.

src/pathfora.js
```diff
@@ -1750,23 +1749,32 @@
...
-          w.content = {
-            0: {
-              title: content.title,
-              description: content.description,
-              url: 'http://' + content.url,
-              image: content.primary_image
+      var recContent = function (w) {
+        pathfora.addCallback(function () {
+          if (pathfora.acctid === '') {
+            pathfora.acctid = context.lio.account.id;
```
markhayden: probably need to make sure that lio and account both exist before assigning. it should be it'll panic if they dont and since we want this to work without lytics, in theory, we should cover all bases.

src/pathfora.js
```diff
@@ -2261,97 +2279,226 @@
...
+        defaultEl.elem.setAttribute('data-pfmodified', 'true');
+        this.preppedElements[group] = defaultEl;
+      }
     };
 
+    this.procRecommendElements = function (blocks, rec, cb) {
+      var inline = this;
+
+      if (rec !== 'default') {
+        // TODO: NEED TO MAKE SURE FILTER IS WORKING, SEEMS BROKEN ATM
```
markhayden: probably dont want this in production :) maybe a note about stability or something....broken = bad :)

src/pathfora.js
```diff
@@ -2416,6 +2569,10 @@
       if (document.addEventListener) {
         document.addEventListener('DOMContentLoaded', function () {
           pf.addCallback(function () {
+            if (pf.acctid === '') {
+              pf.acctid = context.lio.account.id;
```
markhayden: same here. should probably make sure `lio.account` exists and if not throw an error to the console.

src/pathfora.js
```diff
@@ -1750,23 +1749,32 @@
...
-              description: content.description,
-              url: 'http://' + content.url,
-              image: content.primary_image
+      var recContent = function (w) {
+        pathfora.addCallback(function () {
+          if (pathfora.acctid === '') {
+            pathfora.acctid = context.lio.account.id;
+          }
+
+          api.recommendContent(pathfora.acctid, w.recommend.ql.raw, function (resp) {
```
markhayden: not sure if this is the right place for it but some commenting on what exactly will happen / how content recos might be useful (e.g. how defaults work, how many are pulled in, etc)

treat it like another dev is looking through the code outside of lytics to understand whats going on.

src/pathfora.js
```diff
@@ -2236,20 +2247,27 @@
...
+          callback([]);
         }
       }, function () {
-        callback(null, widget);
+        callback([]);
       });
+    },
+
+    /**
+     * @description Construct filter for inline content recommendations
```
markhayden: might link to docs here if we have any, if not we should add a todo line about getting more info there. i know its preliminary so don't sweat it but its where people will get stuck for sure.

src/pathfora.js
sarahyablok: Whoops yeah, this actually isn't even broken anymore. 😁 So I'm just gonna remove this comment.

src/pathfora.js
sarahyablok: Added some comments here and a few examples of valid url patterns. Unfortunately I don't think we have any real public docs on this yet. :/

dist/pathfora.js
```diff
@@ -2621,7 +2651,7 @@
      * @param {object|array}   widgets
      * @param {object}         config
      */
-    this.initializeWidgets = function (widgets, config) {
+    this.initializeWidgets = function (widgets, config, oldConfig) {
```
markhayden: this is fine tho it seems a little clunky, especially if it changes again? another approach might be something like 

```javascript
var sampleargs = function () {
	var widgets, config;

	// in the legacy version of pathfora we allowed for account id to be passed
	// this is now pulled directly from lio to avoid missconfigurations
	if (arguments.length >= 3) {
		widgets = arguments[0];
		config = arguments[2];
	} else {
		widgets = arguments[0];

		if (typeof arguments[1] !== 'string') {
			config = arguments[1];
		}
	}

	console.log('widgets: ' + widgets, '  config: ' +config);
}
```

markhayden:
👍

PR #124, opened 2016-07-18
Dom load fix for inline elements

sarahyablok:
Copied over some code from outlio to fix the same problem with the DOM load listener. Even if the DOM has already loaded we should add the callback with the call to `pathfora.inline.procElements`.


markhayden:
👍  after tests pass

PR #126, opened 2016-07-19
Docs for Inline personalization and content recommendations

sarahyablok:
Adds some basic documentation for inline personalization and inline content recommendations including a new section in the docs. Each page has an explanation of the setup, special attributes, and a detailed example.

I'd really like to get some fake content (or just copy over the content table from the Lytics account) into the Pathfora account. Then in the examples we can make legitimate recommendations. Currently the recommendation examples will just show the default content which is a little underwhelming.

These docs are strictly developer focused and are not nearly as cool/useful as the old squeaky fuzzy. I'm thinking when we have time we can build out better examples for sales/marketing. This is just the bare minimum that we can share with developers.


docs/docs/inline_content.md
```diff
@@ -0,0 +1,84 @@
+Pathfora also supports "fill in the blank" style inline content recommendations. That is, you can set up your html elements to match the look and feel of your website, and add attributes which will signal to pathfora to fill in the content with results from the Lytics content recommendation API.
```
araddon: Is this "fill in the blank" needing a replacement?

araddon:
:+1:

PR #128, opened 2016-08-01
Add optional country field to form

### LINKED ISSUE #127: Add Form Field: Country
Need to add an optional field to form & sitegate modules for Country. This should be a drop down/select list.

sarahyablok:
closes #127 
Used this default list:
http://www.mediacollege.com/internet/samples/html/country-list.html

Can change if necessary!

screen shot 2016-08-01 at 1 54 22 pm

screen shot 2016-08-01 at 1 54 28 pm


src/templates/includes/country.html
```diff
@@ -0,0 +1,250 @@
+<select name="country">
+  <option value="">Country</option>
+  <option value="Afganistan">Afghanistan</option>
```
markhayden: it looked like on the site u pulled these from they had 2 options, one for full name as value and another abbreviated one. might be best to use abbreviated as it makes for cleaner data? probably a preference thing tho. full name is good for the actual text in the select tho.

markhayden:
👍

PR #130, opened 2016-08-23
use lytics-js-styleguide

humanchimp:
Hello. This (possibly controversial) PR modifies this repo to depend on and follow the ESLint configuration included in the [new Lytics JS styleguide](https://github.com/lytics/lytics-js-styleguide). The goal is to converge on a set of ESLint rules that can be used for all Lytics JavaScript projects. The rules in lytics-js-styleguide are not etched in stone, but they are reasonable, and they are the ones that we're using in the admin app.


src/pathfora.js
```diff
@@ -142,7 +141,7 @@
   // NOTE HTML templates
   // FUTURE Move to separate files and concat
   /* eslint-disable indent */
-  var templates = {{templates}};
+  var templates = '{{templates}}';
```
sarahyablok: I see why you did this, because when I was doing my linting PR I did the same thing, but we decided against it. The problem is we don't really want to be finding/replacing our templates differently like this. Sadly there's no good way to disable this causing an error and @markhayden and I later agreed to only lint the `dist` version. I think if you run our lint task `gulp lint` as is, it will only check the dist file.

Idk if either of these are the best solution? Maybe it's something we should revisit? @markhayden any thoughts?

.eslintrc
```diff
@@ -1,121 +1,10 @@
...
+extends:
+- ./node_modules/lytics-js-styleguide/.eslintrc
+env:
+  browser: true
+  jasmine: true
+globals:
+  pathfora: true
+  "$": true
+  FB: true
+  gapi: true
```
sarahyablok: Would it be acceptable to add any pathfora-specific rules here? For example the enforcement of single quotes. Since it's already everywhere in the code it would be nice to maintain. I don't think we would plan on adding many rules back, but a few one-off things like that might be nice to keep.

sarahyablok:
If you haven't already, I'd try running the `gulp lint` task. It will check all the js for the docs as well. That's the task we'd like to encourage people using for linting in pathfora.

.eslintrc
humanchimp: Indeed. I think for eminently reasonable rules like enforcing single quotes (unless it requires backwhacking), we should incorporate them into https://github.com/lytics/lytics-js-styleguide. Also, it's totally acceptable to override things locally.

humanchimp:
Pretty sure this is passing lint as-is. Let me verify.

sarahyablok:
Ah, you're right it is passing, but I wonder why it's not checking `docs/docs/examples/**/*.js`? I'll investigate.

humanchimp:
> It is checking those files, I guess there was nothing to change?

It was my opinion that some of the rules in https://github.com/lytics/lytics-js-styleguide were too quaint/strict to be enforced in documentation, so I relaxed them using a [local .eslintrc file](https://github.com/lytics/pathforajs/blob/62392fd9fba75e919eb56396dae571ef6c6386b5/docs/.eslintrc).

src/pathfora.js
humanchimp: One idea: we can use a syntactic macro. Instead of having handlebars, we'd just call a function
```javascript
var templates = HBS('templates');
```
```javascript
function HBS() {
  return ...; // Maybe this returns fixture data, or it might just be a noop
}
```
but at compile time we could inline the actual templates and eliminate the `HBS` function using dead code removal

.eslintrc
humanchimp: I updated https://github.com/lytics/lytics-js-styleguide to include the single quotes rule, and bumped the version we're depending on here.

sarahyablok:
Besides the templating thing 👍 Maybe @markhayden, you, and I should have a quick brainstorm on that?

sarahyablok:
Unfortunately I think we're going to close this one. (sorry it looks like it never got a second review). I am going to add a bullet item to the future pathfora roadmap ticket about re-assessing our lint rules. That may mean using the lytics-js-styleguide if it is still up to date.

PR #133, opened 2016-09-06
concat widget lists

markhayden:
when leveraging a config with multiple widgets targeting multiple segments we were just overwriting with each loop and thus only ever showing the last widget in the array (if it matched). this pr ensures that we concatenate the array so that all widgets with proper segment matches have the ability to show.


DingoEatingFuzz:
Tests?

markhayden:
si, tests added

test/pathforaSpec.js
```diff
@@ -476,6 +476,56 @@ describe('Pathfora', function () {
...
+
+    expect(widgetA).toBeDefined();
+    expect(widgetB).toBeDefined();
+
+    setTimeout(function () {
+      expect(widgetA.hasClass('opened')).toBeTruthy();
+      expect(widgetB.hasClass('opened')).toBeTruthy();
+      pathfora.clearAll();
+      done();
+    }, 200);
```
markhayden: there wasn't originally but i believe there might be now. its done like this in lots of places tho so going to make a ticket and fix them all at once.

DingoEatingFuzz: 👍

DingoEatingFuzz:
:shipit:

PR #135, opened 2016-09-07
Support new Recommendation Params

### LINKED ISSUE #131: Support new recommendation filters
Content recommendation modal configs can now be expanded to support the following new fields:

- `topics`  - topics to which the recommendations should be constrained
- `rollups`  -rollup topics to which the recommendations should be constrained
- `from`  - start date (published) of recommendations to use
- `to`  - end date (published) of recommendations to use
- `path`  - url path of documents to filter
- `domain` - domain of acceptable documents to return
- `rank`  - method of ranking to use, currently {popular|affinity|recent}
- `visited` - whether to show urls that the user has seen

we should maintain support for raw ql as well.

To start I will add support for these filters for modals only, we should rethink the way we support inline recommendations or possibly wait for content collections.

sarahyablok:
closes #131 

- Supports all the new recommendation filters that are supported by the API query params described in the ticket above. Adds documentation for these new config variables.

- ~~Also we are doing a quick change on the `branding` control setting before we had it controlled with a boolean, now we're switching to `branding: "show"` and `branding: "hide"` to account for some go/json funkyness for native campaign overrides.~~ Reverted this change as we thought of a better way to handle this on the back-end.

~~Should probably add some additional tests to make sure the params are before constructed correctly in all cases before merging this.~~


dist/pathfora.js
```diff
@@ -143,6 +143,7 @@
   // FUTURE Move to separate files and concat
   /* eslint-disable indent */
   var templates = {
+  'templates': {},
```
markhayden: should this little fella be here?

src/pathfora.js
```diff
@@ -2264,24 +2264,60 @@
...
         '{{apiurl}}/api/content/recommend/',
         accountId,
         '/user/_uids/',
-        seerId,
-        filter ? '?ql=' + filter : ''
-      ].join('');
+        seerId
+      ];
+
+      // construct the query params
```
markhayden: it probably make sense to break this out into a utility function so it can be used elsewhere as we expand / it would be easier to write tests around it

dist/pathfora.js
sarahyablok: It's strange. I've noticed that when I run the gulp tasks that gets added. It might be an extra lurking folder on my local machine or something, will investigate.

markhayden: cool, not a deal breaker. we can always open another ticket to figure it out. not going to hurt n e thing.

markhayden:
👍

PR #139, opened 2016-10-14
Account for theme value of 'none'

sarahyablok:
For Lytics campaigns, the `theme` value in the config for custom CSS modules is now `none`.


markhayden:
👍

PR #141, opened 2016-10-29
New exit intent display option

DingoEatingFuzz:
Barebones implementation, looks like this:

``` js
var subscription = pathfora.Message({
  layout: "modal",
  id: 'modal-subscription',
  headline: 'Don\'t leave yet!',
  msg: 'Please, anything but that.',
  theme: 'dark',
  okMessage: 'Sure, whatever',
  okShow: true,
  displayConditions: {
    showOnExitIntent: true // :o
  }
});
```


sarahyablok:
Something to consider: In your current example I can see the modal the first time when I "attempt to exit" the window by moving my cursor towards the top of the browser. Lets say I read the modal, and I'm still not interested in what it has to say, but I go and take the time to hit the "x" button to make it go away, cause I don't like pop-ups. I go back to try and exit the window, and there it is again! Even though I just took the time to close it.

I don't think we should ever show the exit intent modal more than once per page view. You might think this could be remedied by coupling this with `hideAfterAction` rules, but sadly it cannot. The following settings:

``` javascript
showOnExitIntent: true,
  hideAfterAction: {
    closed: {
      hideCount: 1,
    }
  }
}
```
... result in the modal being shown any number of times in the first page view given the conditions above, but on refresh it will never show again. Let's consider setting some kind of default with this setting so that we don't display it multiple times per page view.

DingoEatingFuzz:
> I don't think we should ever show the exit intent modal more than once per page view.

That's fair.

PR #142, opened 2016-11-08
Fix acctid bug for inline content recommendations

sarahyablok:
Encountered a bug where pathfora couldn't tried referencing itself to get the account ID `pathfora.acctid` before it had finished initializing the pathfora object. This should fix it, since now we pass the acctid from the jstag to the inline object.

Also decided to add author and date fields to the inline content recommendations. For date formatting, the user can set options and a locale in js:

``` javascript
pathfora.locale = "en-US";
pathfora.dateOptions = { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' };
```


dist/pathfora.js
```diff
@@ -2555,7 +2557,7 @@
 
             // CASE: Content recommendation elements
             case 'data-pfrecommend':
-              if (context.pathfora.acctid === '') {
+              if (pathfora.acctid === '') {
```
markhayden: probably should also test if pathfora.acctid is undefined here as well. if for some reason pathfora is not an object pathfora.acctid will be undefined it wont throw an error. i think. 

granted, edge case since we are invoking the function but /shrug

PR #146, opened 2016-11-22
Add additional form event tracking.

### LINKED ISSUE #144: Form Progress Event(s)
There are some valuable use cases that are powered by targeting users that have began to fill out a lead capture but then have abandoned. Lets explore a few ways that we can emit meaningful events to track progress.

For V1 lets keep it simple by emitting a new event as soon as any user controlled field contains > 0 characters. This would show they had intent but would ignore the amount of progress they made (e.g. 2 out of 3 fields) which is probably irrelevant for most initial use cases. 

Likewise, it might be interesting to track hover states as well. Users that have moused over the "submit" or "CTA" buttons but did not click, might be an interesting way to build an audience of users likely to convert?

sarahyablok:
closes #144

reference: https://github.com/lytics/pathforajs/issues/144#issuecomment-261671082


dist/pathfora.js
```diff
@@ -1053,10 +1047,11 @@
...
      */
-    removeWatcher: function (watcher) {
-      for (var key in core.watchers) {
-        if (core.watchers.hasOwnProperty(key) && watcher === core.watchers[key]) {
-          core.watchers.splice(key, 1);
+    removeWatcher: function (watcher, widget) {
+      for (var key in widget.watchers) {
+        if (widget.watchers.hasOwnProperty(key) && watcher === widget.watchers[key]) {
+          widget.watchers.splice(key, 1);
+          console.log(key);
```
markhayden: probably wanna kill this log

PR #160, opened 2016-12-22
Exclude UrlContains rules

### LINKED ISSUE #152: Support exclusion rules in the style of urlContains
Sometimes it's desired to show a widget on all pages _except_ A, B, and C. In these cases, it's best to blacklist instead of whitelist. 

Potential API ideas:

```js
{
  displayConditions: {
    urlContains: [
      {
        match: 'simple',
        value: '/show-here'
      },
      {
        match: 'simple',
        value: '/dont-show-here',
        exclude: true
      },
    ]
  }
}
```

or

```js
{
  displayConditions: {
    urlContains: [
      {
        match: 'simple',
        value: '/show-here'
      },
    ],
    urlExcludes: [
      {
        match: 'simple',
        value: '/dont-show-here'
      },
    ]
  }
}

### PR DESCRIPTION
**Pull request by sarahyablok**
closes #152

### CODE REVIEW

**Comment by markhayden** [comment]
Need to update this one so that logic works as follows:

```
{
  match: 'string',
  value: 'examples'
},
{
  match: 'string',
  value: 'notreal'
},
{
  match: 'string',
  value: 'supernotreal',
  exclude: true
},
{
  match: 'string',
  value: 'somethingelse',
  exclude: true
}
```

translates to

```
(
  {
    match: 'string',
    value: 'examples'
  }
  OR
  {
    match: 'string',
    value: 'notreal'
  }
)
AND
(
  {
    match: 'string',
    value: 'supernotreal',
    exclude: true
  }
  OR
  {
    match: 'string',
    value: 'somethingelse',
    exclude: true
  }
)
```

so that any matched exclude rule overrides any matched include rule.

PR #162, opened 2016-12-23
Add referrer form field

### LINKED ISSUE #161: Referrer Form Field
Add a form input field for referring a friend. (email)

sarahyablok:
closes #161 

Also make the half-width input fields collapsible.


src/pathfora.js
```diff
@@ -88,7 +88,8 @@
         message: 'Message',
         company: 'Company',
         phone: 'Phone Number',
-        country: 'Country'
+        country: 'Country',
+        referrer: 'Referrer'
```
markhayden: should this be `referral` or even `referral-email`? i think technically in this case the referrer and the main email would be the same (e.g. mark hayden filled out the form and he referred sarah...which makes mark the referrer and sarah the referent....words are hard. just want it to be clear that the 1) its an email and 2) it doesn't belong to the person submitting the data.

also, i think we talked about doing some custom js for this upfront but we might be able to pull it off with just lql. need to run it by aaron to be sure but pretty sure we can just add another lql file that reads from default. if referral shows up we can add that user and ignore the uid and all that stuff....i believe.

markhayden: this was a bad place to add this comment, to be clear i was speaking to the actually key not just the label. the label should definitely be "Referral Email" or something to that effect.

src/pathfora.js
```diff
@@ -1232,27 +1236,33 @@
...
+          var element = getFormElement(field),
+              parent = element.parentNode;
+
+          if (element && !config.fields[field] && parent) {
+            parent.removeChild(element);
+          }
+        });
+
+        // NOTE: collapse half-width inputs
+        Array.prototype.slice.call(widget.querySelectorAll('form .pf-field-half-width')).forEach(function (element, halfcount) {
```
DingoEatingFuzz: `forEach` doesn't work in IE < 9. Not sure what support we are shooting for, but this is easy enough to turn into a loop for compatibility.

DingoEatingFuzz: Well I guess we are already using `forEach` elsewhere, so ignore me.

PR #164, opened 2016-12-28
Trigger Widget Via Javascript

### LINKED ISSUE #156: Trigger Widget via JavaScript
There have been a few requests, plus it seems pretty useful in the long run, to be able to trigger a modal manually. Meaning, instead of using the built in "after 3 seconds" type stuff, we listen for a particular call to Pathfora before a widget can load. 

I'd imagine we'd need to treat it like a scroll listener or any of the listeners because each widget may use different types of triggers. 

For v1 we can have a single call that gives the ok to any widgets waiting on it, no need to get as granular as per campaign but definitely dont want to prevent the rendering of widgets that are not using that trigger from loading normally.

sarahyablok:
closes #156 

```javascript
// display condition
displayConditions: {
    manualTrigger: true
}
```

```javascript
// example to trigger module
$('my-div').hover(function() {
  pathfora.triggerWidgets(); // trigger all widgets with manualTrigger displayCondition
  // pathfora.triggerWidgets(['widget-id']);  
});
```

I am not tied to the name of any of these methods/settings. Feel free to comment with suggestions for a clearer naming convention if necessary.


src/pathfora.js
```diff
@@ -572,6 +573,16 @@
         core.initializeScrollWatchers(widget);
       }
 
+      if (condition.customTrigger) {
```
markhayden: kind of a nit but i'd call this manualTrigger rather than custom probably. that was its clear that it relies on manual client side js.

src/pathfora.js
```diff
@@ -572,6 +573,16 @@
...
+        watcher = core.registerCustomTriggerWatcher(condition.customTrigger, widget);
```
markhayden: same here, `manualTriggerWatcher`

src/pathfora.js
```diff
@@ -694,6 +705,40 @@
     },
 
     /**
+     * @description check if a customTrigger widget
+     * is ready to be displayed, and if so display the widget
+     * @param {object} widget
+     */
+    triggerWidget: function (widget) {
```
markhayden: naming is probably fine here but if you are looking to change it could use something like `execute` or `evaluate` ... even though its a big wordy at least triggerWidget is accurate and hard to confuse with any of the other functions we have.

src/pathfora.js
```diff
@@ -694,6 +705,40 @@
...
+      var valid;
+
+      for (var key in widget.watchers) {
```
markhayden: dont worry about it for this pr but seems like some of this global validation logic could be moved to its own function to simplify adding in additional triggers down the road? i haven't looked at the other functions so ignore if its not repeated and if it is lets just make a ticket and address later as a whole.

src/pathfora.js
```diff
@@ -982,6 +1027,28 @@
       return watcher;
     },
 
+
+    /**
+     * @description Register a scroll position-triggered widget
```
markhayden: think this comment is copy pasta leftover

src/pathfora.js
sarahyablok: It is. I think while I was making this PR I was planning to pull it out as its own function. I can't remember why I didn't... if there's a reason, or if I forgot. I can check again though.

markhayden: cool, lets just make a ticket and address it down the road. not going to effect n e thing in the interim or prevent backwards compatibility at all.

sarahyablok: Made a ticket: https://github.com/lytics/pathforajs/issues/165

DingoEatingFuzz:
Maybe it's too late to be having this conversation, but why make `manualTrigger` a display condition at all? Why not allow any widget to be triggered manually?

There may be situations where someone wants to show the widget on init in addition to on a trigger, and in the situations where it isn't desired to show the widget on init, there is the existing `[showOnInit](https://github.com/lytics/pathforajs/blob/develop/src/pathfora.js#L51)` display condition.

Then the API surface area would look something like this:

```js
pf.triggerWidget(widget) // show the widget only if display conditions are met
pf.forceTriggerWidget(widget) // show the widget no matter what
pf.triggerAllWidgets() // show all widgets if they each meet their display conditions
pf.forceTriggerAllWidgets() // show all widgets no matter what
```

src/pathfora.js
```diff
@@ -694,6 +705,40 @@
...
+        context.pathfora.showWidget(widget);
+        widget.valid = false;
+        context.pathfora.triggeredWidgets[widget.id] = false;
+
+        // remove from the ready widgets list
+        core.readyWidgets.forEach(function (w, i) {
+          if (w.id === widget.id) {
+            core.readyWidgets.splice(i, 1);
+          }
+        });
```
DingoEatingFuzz: Splicing the array you are `forEach`ing over like this is prone to bugs, since the indices of elements are going to change.

Fortunately there will only ever be one widget to splice, so you could instead use `Array#some` to exit early.

```js
core.readyWidgets.some(function (w, i) {
  if (w.id === widget.id) {
    core.readyWidgets.splice(i, 1);
    return true;
  }
});
```

src/pathfora.js
```diff
@@ -2894,6 +2967,39 @@
...
+     * @description public method to trigger widgets
+     * with the manualTrigger display condition
+     * @param {array}   widgetsIds (optional)
+     */
+    this.triggerWidgets = function (widgetIds) {
+      var pf = this;
+
+      // no widget ids provided, trigger all ready widgets
+      if (typeof widgetIds === 'undefined') {
+        pf.triggeredWidgets['*'] = true;
```
DingoEatingFuzz: Does this ever get flipped back to `false`? This type of one-time state management makes pathfora especially under-equipped for long-running web pages that will need lifecycle control.

test/pathforaSpec.js
```diff
@@ -2195,6 +2195,49 @@ describe('Widgets', function () {
   //  DISPLAY CONDITIONS
   // -------------------------
 
+  it('should show when all manualTrigger widgets are triggered', function () {
+    $(document.body).append('<div class=\'hover-elem\'>Test</div>');
+
+    $('.hover-elem').hover(function () {
+      pathfora.triggerWidgets();
+    });
```
DingoEatingFuzz: Involving the DOM in this test seems unnecessary to me. Why not simply call `pathfora.triggerWidgets()` below instead of triggering the mouse events?

src/pathfora.js
sarahyablok: No it doesn't get set back to false. Since it should trigger all widgets with the manualTrigger display condition, if a widget gets initialized after this function is called, we still need some kind of indicator that it should be triggered immediately. What are your thoughts on this?

DingoEatingFuzz: My thoughts on triggering widgets that have `manualTrigger` set, but aren't ready by this time?

¯\\_(ツ)_/¯

PR #167, opened 2017-01-18
Add right and left positions for slideouts

### LINKED ISSUE #151: Additional Positions for Slide Out
A customer has requested top left / right positions for the slide out widgets. More details outlined here https://github.com/lytics/customerprojects/issues/46. 

Should be a fairly simple core update to support those positions since they are already supported by the button widgets.

sarahyablok:
closes #151


src/pathfora.js
```diff
@@ -2871,7 +2871,7 @@
      * @public
      * @description Current version
      */
-    this.version = '0.0.15';
+    this.version = '0.0.16';
```
markhayden: we probably only want to bump the version when we do an update to master from develop, otherwise if multiple branches are going the version number will cause conflicts and weirdness.

PR #171, opened 2017-02-06
Content Recommendations - Support AST

### LINKED ISSUE #170: Content Recommendation AST Support
In addition to segment slug and ql for the recommendation collection we need to support a string of AST. Api support will be added in support of this change with https://github.com/lytics/lio/issues/9064. 

Should be able to just add ast to the existing payload and pass that as a param just like ql. In the case where any combination of the three exist the following priority should be taken:

1) Slug
2) AST
3) QL

No idea what would happen if you use a different definition for all three on a single recommendation but that seems bad :)

sarahyablok:
closes #170 

@DingoEatingFuzz: see the vague docs I wrote below

screen shot 2017-02-06 at 2 03 53 pm


src/pathfora.js
```diff
@@ -2546,20 +2546,40 @@
...
+            queries += 'ql=' + ql.raw;
+
+          // Segment JSON (usually segment AST)
+          } else {
+
+            // convert args back to array
+            if (ast.args) {
+              ast.args = Object.keys(ast.args).map(function (val) {
+                return ast.args[val];
+              });
```
DingoEatingFuzz: Why is `ast.args` ever an object?

docs/docs/content_recommend.md
```diff
@@ -102,11 +108,36 @@ Recommend controls any parameters necessary for making a content suggestions to
...
+      <th>Key</th>
+      <th>Type</th>
+      <th>Behavior</th>
+    </tr>
+  </thead>
+  
+  <tr>
+    <td>args</td>
+    <td>array</td>
+    <td>list of arguments to be evaluated with the operator, these arguments may be identifiers, literals, or further expressions*</td>
```
DingoEatingFuzz: "operator. These" to prevent a run-on sentence. Other than that nit-pick these are successfully vague yet informative 👍 

Hopefully soon we will have a destination page on the AST spec.

src/pathfora.js
sarahyablok: Instead of doing a backwards conversions I just went back and meddled with the code that converts arrays into objects during the config merge and made it stop doing that.

PR #178, opened 2017-02-16
Form Builder v1

### LINKED ISSUE #169: Support for More Advanced Form Building
Its a pretty common theme / request to support alternate inputs and/or input types within our personalization widgets. there are a few things to consider when determining what the solution is here:

1) Flexibility. We want to make sure we support the most common input types (text input, text area, select, button, radio, etc). We can likely get away with rolling this out at separate time as opposed to one big push but definitely worth considering all the options. 

2) Organization. Users are somewhat used to dragging and dropping or moving fields around. Right now our forms are in a fixed order. Need to figure out a simple set of requirements for positioning. 

3) Reporting. Everything done with Pathfora should roll neatly into Lytics. This might be as simple as using an existing map for all new fields or something to that effect but we definitely want to know what this looks like through the lens of the Lytics dashboard before moving forward. 

4) Third party vendors. One thing we can explore is integrating tools. I personally think that we should support form building natively. Our styles are already pretty flexible as far as size and positioning. It really just comes down to the config. Third party tools require another step and their own set of potential challenges. Either way we should explore all options before diving in head first. 

All that said lets first come up with a definitive list of requirements. @sarahyablok I think you and I can probably chat for 30 minutes this week and hammer out that based on what customers have asked / what is coming down the pipe. Then we can branch off and figure out the best approach.

sarahyablok:
closes #169 

(details tracked in the ticket above)


araddon:
great job, thx

src/pathfora.js
```diff
@@ -1296,77 +1439,105 @@
...
-            if (element.className.indexOf('pf-field-half-width') !== -1) {
+            if (element && config.required[field]) {
+              element.setAttribute('data-required', 'true');
+            }
+          });
+
+          // Hide fields
+          Object.keys(config.fields).forEach(function (field) {
+            var element = getFormElement(field),
+                parent = element.parentNode;
```
markhayden: this will throw an error if the element is not found (its happening in the existing legacy form example because the key is wrong `referrer` should be `referralEmail`. regardless, we probably want to gracefully handle the issue here by checking for a value and throwing an error if its null or whatever.

sarahyablok:
@markhayden Thanks for pointing that CSS bug out. I suspect it's been lingering for a while now, but we've never tested with modals long enough to notice. This should be fixed as part of the newest commit.

PR #183, opened 2017-03-08
Support content recommendations for inline js widgets

### LINKED ISSUE #182: Inline (layout) modules support content recommendations
Currently our [content recommendation modules](http://lytics.github.io/pathforadocs/content_recommend/) are only accepted for modal and slideout layouts. Just need to make a quick change to allow them with inline modules. We already support inline content recommendations via HTML: http://lytics.github.io/pathforadocs/inline_content/ but for those who want to use modules/js I will add the option.

sarahyablok:
closes #182


src/pathfora.js
```diff
@@ -3593,6 +3607,17 @@
           context.pathfora.closeWidget(widget.id, true);
         }, widget.displayConditions.hideAfter * 1000);
       }
+
+      core.widgetResizeListener(widget, node);
+
+      var resize = window.onresize;
+      window.onresize = function (event) {
+        if (typeof resize === 'function') {
```
markhayden: this is going to always evaluate to true. were you going for a skip the resize if resize was undefined before hitting like 3613?

sarahyablok: I don't think so. The idea was to preserve the `window.onresize` function if it had already been set by the user elsewhere in javascript. So if it hasn't `resize` will be `null`. And we skip calling `resize(event)`. I tested it and I think that's how it should work. It does look kind of strange though...

examples/inline-widgets.html
```diff
@@ -26,7 +26,7 @@
     <div class="inline-widget-host"></div>
     <div class="inline-widget-host2"></div>
 
-    <script src="https://api.lytics.io/api/tag/0d47e6304d12858905b4f111090a68a1/lio.js"></script>
+     <script src="https://api.lytics.io/api/tag/0d47e6304d12858905b4f111090a68a1/lio.js"></script>
```
sarahyablok: 😢

PR #185, opened 2017-03-10
Ensure content recommendations are tracked

sarahyablok:
Ensure that the content data is being tracked when a recommendation widget is shown, closed, or clicked on.


dist/pathfora.js
```diff
@@ -3191,7 +3214,7 @@
      * @public
      * @description Current version
      */
-    this.version = '0.0.16';
+    this.version = '0.1.0';
```
sarahyablok: Ignore this, it should have been out with the last release. I forgot to run gulp after I bumped the version. :|

dist/pathfora.js
```diff
@@ -2222,8 +2239,12 @@
         pathforaDataObject.closedWidgets.push(params);
         break;
       case 'confirm':
-        params['pf-widget-action'] = !!widget.confirmAction && widget.confirmAction.name || 'default confirm';
-        pathforaDataObject.completedActions.push(params);
+        if (htmlElement && utils.hasClass(htmlElement, 'pf-content-unit')) {
+          params['pf-widget-action'] = 'content recommendation';
```
markhayden: do we want spaces in this? probably fine? do we have them in any of the others?

sarahyablok: Yeah, that's the only reason I put spaces in it, cause the others have them. :/ Kinda weird, but at least it's consistent.

PR #188, opened 2017-03-13
Implement Entity Specific Values in Content

### LINKED ISSUE #137: Entity Specific Values in Content
We already have the ability through Lytics to pass a white listed field to pathfora (e.g. email address). There is, however, no simple way to get a dynamic value into the body or title of a pathfora widget. For instance, a promo code for a specific set of users. 

Lets explore a simple template solution, similar to handlebars perhaps, that allows for basic string replacement on a pathfora widget. We need to take care to think of how to handle default values or possibly even formatting here as well. 

example:

```
Use promo code: {{user_promo | DEFAULTPROMO}} today!
```

where defaultpromo is the default string. likewise, we could always take the approach of "if there is no key for that user throw it out" and prevent the modal from being loaded for anyone that doesn't meed the full specs?

sarahyablok:
closes #137


docs/docs/customization/entity_fields.md
```diff
@@ -0,0 +1,37 @@
...
+  id: 'welcome-back',
+  layout: 'modal',
+  headline: 'Welcome back, {{name}}.',
+  msg: 'Thanks for returning. Please check out our new products.'
+});
+
+pathfora.initializeWidgets([module]);
+```
+
+Pathfora will attempt to replace `{{name}}` with the value of `window.lio.data.name`. Similarly, if you wanted to include a subfield, you can use a dot notation in the template. `{{last_cta.resource}}` would be replaced with the value of `window.lio.data.last_cta.resource`.
```
markhayden: `you can use a dot notation` => `you can use dot notation`

docs/docs/customization/entity_fields.md
```diff
@@ -0,0 +1,37 @@
...
+
+If the data isn't present for the user viewing the modal there are a couple options available to us to control the behavior of the module. By default if there is no fallback defined for the field, pathfora will not display any module including data does not exist for the user. This behavior can be changed with the [showOnMissingFields](/display_conditions.md#showonmissingfields) display condition.
```
markhayden: `options available to us to control the behavior` => `options available to control the behavior`

markhayden: `By default if there is no fallback defined for the field, pathfora will not display any module including data does not exist for the user.` => `If there is no fallback defined for the field, pathfora will prevent the module from being displayed for visitors who lack the necessary fields.`

docs/docs/customization/entity_fields.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Defining a Fallback (default) Value
+
+Alternatively, we can define a fallback or "default" value to be shown in case the user is missing the data. To define a fallback value, place it after the field name and a `|` character in the template. Or, continuing the example from above:
```
markhayden: ` template. Or, continuing the example from above:` => ` template. Building from the previous example:`

docs/docs/customization/entity_fields.md
```diff
@@ -0,0 +1,37 @@
...
+  id: 'welcome-back',
+  layout: 'modal',
+  headline: 'Welcome back, {{name | Valued Customer}}.',
+  msg: 'Thanks for returning. Please check out our new products.'
+});
+
+pathfora.initializeWidgets([module]);
+```
+
+So if a user whose name is unknown visited a website with this module, they would be greeted as a "Valued Customer".
```
markhayden: `So if a user whose name is unknown visited a website with this module, they would be greeted as a "Valued Customer".` => `In this example all visitors who lack the "name" field will be greeted with "Welcome Back Valued Customer".`

src/pathfora.js
```diff
@@ -866,6 +895,53 @@
...
 
+    entityFieldChecker: function (widget, fieldName, found) {
+      if (!found || !found.length) {
+        return true;
+      }
+
+      // for each template found...
+      for (var f = 0; f < found.length; f++) {
+        // parse the field name
+        var dataval = found[f].slice(2).slice(0, -2),
```
markhayden: this is fine but might be safer / support more complex things if we use regex here to define the parts rather than splits.

src/pathfora.js
```diff
@@ -866,6 +895,53 @@
...
+            parts = dataval.split('|'),
+            def;
+
+        // get the default (fallback) value
+        if (parts.length > 1) {
+          def = parts[1].trim();
+        }
+
+        // check for subfields if the value is an object
+        var split = parts[0].trim().split('.');
```
markhayden: regex can also prevent the potential situation where a period was in the fallback and it was intended to be punctuation rather than dot notation. (e.g. if a period has no spaces before or after it, can be considered to be dot but if it has a space it could be something like `Dr.`

sarahyablok: Well, right now it's only doing this split on the fieldname, so it doesn't even look for dots in the fallback, but I see your point.

markhayden: i think it really only matters if we start to support more advanced logic, edge case for sure.

PR #189, opened 2017-03-16
Store recommendation responses in sessionStorage

### LINKED ISSUE #187: Keep Recommendation Response in localStorage
Talked to @drewlanenga last week, and it was suggested that we keep recommendation response in localStorage for a set limit of time (I'm thinking a single session). To reduce the number of requests we're making to the content recommendation API.

Some things to keep in mind:

- displaying new articles each time from a `shuffle` param request
- how do we ensure that the article is refreshed if they visit it and the `visited` param is on?
- having a higher limit (the default right now is 20) on the recommendation request ensures more relevant recommendations since it allows for a larger sample size, but may result in longer response times.

The first two we talked about possible solutions for, and I will revisit them when I begin working on this.

sarahyablok:
We can switch this over to localStorage and I'll implement some arbitrary time to make it expire if we still think that once per session is still too much, the response is too large to store in a session cookie.

closes #187


src/pathfora.js
```diff
@@ -2824,10 +2824,32 @@
...
+    recommendContent: function (accountId, params, id, callback) {
       // Recommendation API:
       // https://www.getlytics.com/developers/rest-api#content-recommendation
 
+      // if we have the recommendation response cached in session storage
+      // use that instead of making a new API request
+      var storedRec = sessionStorage.getItem('PathforaRecommend_' + id);
+
+      if (typeof storedRec === 'string' && params.visited !== false) {
+        var rec = JSON.parse(storedRec);
```
markhayden: probably should try/catch here to be safe. a plain string will blow up with JSON.parse if its not actually JSON. should be fine since its our api but im thinking we want this as bulletproof as possible?

src/pathfora.js
```diff
@@ -2824,10 +2824,32 @@
...
+
+        if (rec.data) {
+          // special case: shuffle param
+          if (params.shuffle === true) {
+            rec.data.shift();
+
```
markhayden: extra line

src/pathfora.js
```diff
@@ -2824,10 +2824,32 @@
...
      * @param {string} accountId  Lytics account ID
      */
-    recommendContent: function (accountId, params, callback) {
+    recommendContent: function (accountId, params, id, callback) {
       // Recommendation API:
       // https://www.getlytics.com/developers/rest-api#content-recommendation
 
+      // if we have the recommendation response cached in session storage
+      // use that instead of making a new API request
+      var storedRec = sessionStorage.getItem('PathforaRecommend_' + id);
```
markhayden: we should probably store the recommend prefix as a constant and reference that constant throughout the library.

src/pathfora.js
```diff
@@ -2824,10 +2824,32 @@
...
+
+        if (rec.data) {
+          // special case: shuffle param
+          if (params.shuffle === true) {
+            rec.data.shift();
+
+          }
+
+          if (rec.data.length > 0) {
+            sessionStorage.setItem('PathforaRecommend_' + id, JSON.stringify(rec.data));
```
markhayden: same constant as above

src/pathfora.js
```diff
@@ -2874,6 +2896,9 @@
       var recommendUrl = recommendParts.join('') + queries;
 
       this.getData(recommendUrl, function (json) {
+        // set the session storage.
+        sessionStorage.setItem('PathforaRecommend_' + id, json);
```
markhayden: same constant

src/pathfora.js
sarahyablok: 👍 Yeah, we have a bunch of these throughout the library I'll go ahead and pick them out.

src/pathfora.js
```diff
@@ -2830,20 +2838,25 @@
...
-        var rec = JSON.parse(storedRec);
+        var rec;
+
+        try {
+          rec = JSON.parse(storedRec);
+        } catch (e) {
+          console.warn('Could not parse json stored response:' + e);
+        }
 
         if (rec.data) {
```
markhayden: need to: `if(rec && rec.data){` here if we are going for bulletproof. in the event the json parse fails rec would be undefined and this would throw an error.

```
var rec;
try {
  rec = JSON.prase('invalid');
} catch (e) {
  console.warn('Could not parse json stored response:' + e);
}

console.log(rec.data);
```

PR #190, opened 2017-03-17
Check for Relative URLs from recommendation API

### LINKED ISSUE #184: Qualify relative URL paths before rendering content recommendations
Currently, URLs with relative paths (`path/to/content`) are prefixed with an absolute path prefix (`//path/to/content`) .  We'll probably either want to remove absolute path prefix (`//`), or prepend the domain (`//domain.com/path/to/content`) before rendering the recommendation widget.

sarahyablok:
closes #184


src/pathfora.js
```diff
@@ -2875,7 +2875,19 @@
 
       this.getData(recommendUrl, function (json) {
         var resp = JSON.parse(json);
```
markhayden: considering this is our api it shouldn't be an issue but might be worth doing a try/catch here to handle the edge case of invalid json. at the moment if that api returns invalid json it will throw an error. seems like in that case we'd want to raise a warning and then return `[]`

dist/pathfora.js
```diff
@@ -2910,7 +2910,19 @@
...
         var resp = JSON.parse(json);
+
         if (resp.data && resp.data.length > 0) {
+          // append a protocol for urls that are absolute
+          for (var i = 0; i < resp.data.length; i++) {
+            var url = resp.data[i].url;
+            if (url) {
+              var split = url.split('/')[0].split('.');
+              if (split.length > 1) {
+                resp.data[i].url = 'http://' + url;
```
markhayden: i could be derping but are we sure this will always be insecure? i realize we can't just pull the current protocol and append that for a recommendation that is not necessarily in the same environment but seems like worst case its a ticket we should make related to content recommendations? kinda seems yuck for us to be manipulating urls to something that is being returned from an api? 

to be clear, fine for now but seems like it smells bad.

docs/docs/index.md
```diff
@@ -1,12 +1,9 @@
...
 
 ## Getting Started
-Before you begin creating modules with Pathfora you must add the [Lytics javascript tag](https://activate.getlytics.com/#/documentation/jstag_anon) to your website. If you are using [Lytics Personalization](https://activate.getlytics.com/resources/documentation/campaigns_introduction) to power your modules, Pathfora will be automatically loaded through your Lytics javascript tag. If you're creating a custom module, the Pathfora tag must be loaded after the Lytics tag.
+Before you begin creating modules with Pathfora you must add the [Lytics javascript tag](https://activate.getlytics.com/#/documentation/jstag_anon) to your website. If you are using [Lytics Personalization](https://activate.getlytics.com/resources/documentation/campaigns_introduction) to power your modules, Pathfora will be automatically loaded through your Lytics javascript tag. If you're creating a custom module, you will need to load pathfora as well as [your Lytics javascript tag](https://activate.getlytics.com/data/integrations/lytics_external).
```
markhayden: pretty sure the `/jstag_anon` url is a broken link.
should be `https://activate.getlytics.com/resources/documentation/getting_started_web`

same for the `/lytics_external` link.
should be `https://activate.getlytics.com/resources/documentation/developers_pathfora_sdk?aid=true&redirect=%2Fdata%2Fintegrations%2Flytics_external%3Faid`

dist/pathfora.js
sarahyablok: Yup, as far as I can tell it's always been like that. Since we don't have the protocol in the API, we can't assume that it is secure or the same as the current page. I think we found that http was "safest" in that most websites that use https will set up a redirect. Would be great if we could have the protocol as part of the URL from the API. I believe @drewlanenga & @DingoEatingFuzz have had conversations around that already.

docs/docs/index.md
sarahyablok: Second link you posted doesn't work for me. Are we trying to do a redirect? I assume people without a Lytics account can't view the integrations page, so is this a redirect that accounts for that?

dist/pathfora.js
drewlanenga: I don't remember any resolution from any specific conversations about it, but adding it as a property in the model is a good idea: https://github.com/lytics/lio/issues/9516

docs/docs/index.md
markhayden: whoops, thats my bad. updated with a working link. The docs should all be public, so trying to link to those.

sarahyablok: I disagree with the second link going to the Pathfora SDK doc. I'll just also link it to the getting_started doc for js tag.

markhayden: thats cool. as long as they link somewhere :)

PR #198, opened 2017-04-04
version 0.1.3

sarahyablok:


markhayden:
we should get the fallback bug into this release as well before merging.

PR #203, opened 2017-04-14
Widget Prioritization v1

### LINKED ISSUE #121: Overlapping Widget Handling
When two widgets are in the same position or even on the same screen it would be great to have more control on how to handle this. Default behavior today is to just show them all, if they overlap that is on the designer.

In a future version maybe they could:
- stack / trigger on change so when one closes the other slides up. 
- allow a priority setting and only show one (this would also need a grouping)
- allow a priority setting for only overlapping

sarahyablok:
closes #121 

---
## priority
Module priority helps prevent overlap between multiple modules on a page. If unset, the module will show as normal, regardless of other modules priorities.

**For example**: module A has a priority of 1 and is targeted to show to the audience "High Value Users".

module B has a priority of 0 is targeted to "Returning Users".

If the user is a member of both audiences, only module A will show assuming all of it's other display conditions are met, because module B is lower priority.

<table>
  <thead>
    <tr>
      <td colspan="2" align="center"><code>priority</code> int</td>
    </tr>
    <tr>
      <th>Key</th>
      <th>Type</th>
    </tr>
  </thead>

  <tr>
    <td style="min-width: 97px;">0 – ∞</td>
    <td>modules with a higher value will be prioritized to show instead of modules with lower numbers assuming targetting requirements and all other page-load level display conditions are met</td>
  </tr>
</table>

``` javascript
// example: module is higher priority than any other module with a priority less than 2.

displayConditions: {
  priority: 2
}
```


src/pathfora.js
```diff
@@ -4019,6 +4033,27 @@
       return core.prepareABTest(config);
     };
 
+    this.reinitializePrioritizedWidgets = function () {
+      if (core.prioritizedWidgets.length > 0) {
+        var highest = [core.prioritizedWidgets[0]];
+
+        for (var i = 1; i < core.prioritizedWidgets.length; i++) {
```
sarahyablok: I one-uped you and just used the array [sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort?v=example) method! :)

PR #209, opened 2017-05-10
Convert to modules with rollup

### LINKED ISSUE #200: modularize the codebase using Rollup
[We just used Rollup to modularize the jstag codebase](https://github.com/lytics/jstag/pull/35), and it worked so well, I think we should modularize pathfora too.

sarahyablok:
closes #200 
We should do some intensive testing on this before it's released, for now tests are passing and everything's looking good in the doc examples, but I'll definitely spend more time testing things on this branch.


src/rollup/globals/reset-default-props.js
```diff
@@ -0,0 +1,128 @@
+/** @module pathfora/globals/reset-default-props */
+
```
markhayden: lol i'd never really noticed this but might be worth making a ticket about handling defaults more gracefully. this approach seems like it might smell a little and be super difficult to scale as new features are added. nothing to do here just a note.

src/rollup/inline/inline.js
```diff
@@ -0,0 +1,36 @@
+/** @module pathfora/inline/inline */
+
+import document from '../dom/document';
+import prepElements from './prep-elements';
+import procElements from './proc-elements';
+import procRecommendElements from './proc-recommend-elements';
+import procTriggerElements from './proc-trigger-elements';
+import setDefaultRecommend from './set-default-recommend';
+
```
markhayden: extra space

src/rollup/pathfora.js
```diff
@@ -0,0 +1,104 @@
+/** @module pathfora */
+
+import { PF_VERSION, PF_LOCALE, PF_DATE_OPTIONS, CSS_URL } from './globals/config';
+
```
markhayden: might be worth adding some comments to each section so its clear whats loaded and why. will be helpful when adding new modules in order to keep similar functionality grouped.

src/rollup/utils/utils.js
```diff
@@ -0,0 +1,32 @@
+/** @module pathfora/utils */
+
```
markhayden: same idea on commenting each section up a bit for clarity.

src/rollup/widgets/close-widget.js
```diff
@@ -0,0 +1,49 @@
...
+
+import removeClass from '../utils/class/remove-class';
+import hasClass from '../utils/class/has-class';
+
+import { widgetTracker } from '../globals/config';
+
+
+export default function closeWidget (id, noTrack) {
+  var node = document.getElementById(id);
+  var i;
```
markhayden: nit but seems like in most places we put undefined variables at the top level of the var definition and separate additional definitions with commas rather than multiple var statements. not sure i have a preference, both have benefits and drawbacks, but we should at least be consistent. nit tho for sure.

src/rollup/widgets/form.js
```diff
@@ -0,0 +1,7 @@
+/** @module pathfora/widgets/form */
+
+import prepareWidget from './prepare-widget';
+
+export default function Form (config) {
```
markhayden: might be a little generic on the naming. ur call tho.

src/rollup/widgets/gate.js
```diff
@@ -0,0 +1,7 @@
+/** @module pathfora/widgets/gate */
```
markhayden: module and function name are different. been pretty consistent everywhere else.

src/rollup/data/tracking/get-data.js
```diff
@@ -0,0 +1,7 @@
+/** @module pathfora/data/tracking/get-data */
+
+import { pathforaDataObject } from '../../globals/config';
+
+export default function getData () {
```
sarahyablok: @markhayden yeah, I think I only kept the name here cause it's a public method. But I doubt anyone is using it besides our tests so I don't feel like there's any harm in changing it.

src/rollup/ab-test/init-ab-test.js
```diff
@@ -0,0 +1,53 @@
+/** @module pathfora/ab-test/init-ab-test */
+
+import saveCookie from '../utils/cookies/save-cookie';
+import readCookie from '../utils/cookies/read-cookie';
+
+import { pathforaDataObject } from '../globals/config';
```
sarahyablok: Import order is not an issue in rollup, because yeah it gets imported in the child module as well. I can try to create some sort of consistent order for imports though. Currently things are grouped in the same way that the folders are, which is why there are the random spaces you are seeing. Might make more sense if I just add comments for those groups though.

src/rollup/display-conditions/pageviews/page-visits-checker.js
```diff
@@ -0,0 +1,17 @@
+/** @module pathfora/display-conditions/pageviews/page-visits-checker */
+
+// globals
+import { PF_PAGEVIEWS } from '../../globals/config';
+
+// utils
+import readCookie from '../../utils/cookies/read-cookie';
+
+/**
+ * Check if the pagevist count meets the requirements
```
markhayden: `pagevist` => `pagevisit`

src/rollup/display-conditions/url-contains/url-checker.js
```diff
@@ -0,0 +1,55 @@
...
+      valid, excludeValid = false,
+      matchCt, excludeCt = 0;
+
+  if (!(phrases instanceof Array)) {
+    phrases = Object.keys(phrases).map(function (key) {
+      return phrases[key];
+    });
+  }
+
+  // array of urlContains params is an or list, so if any are true evaluate valid to true
```
markhayden: super nit but might be worth capitalizing `or` here so it doesn't read like `params is an or` ... an or what? :)

PR #216, opened 2017-06-06
Allow for entity field templates in callback functions

sarahyablok:
This is useful for dynamic URLs, specifically in the Lytics campaign builder. For example the user may now input something like:

`http://mywebsite.com/a/page?userid={{userid}}`

Into the target url input, where userid will be replaced with the userid value for that entity in Lytics.


src/rollup/display-conditions/entity-fields/entity-field-checker.js
```diff
@@ -0,0 +1,59 @@
...
+      pf = this,
+      count = 0;
+
+  // call the replace method in a jstag callback
+  var replace = function (w, fieldName, f) {
+    pf.addCallback(function () {
+      w.valid = w.valid && pf.replaceEntityField(w, fieldName, f);
+      count++;
+
+
```
markhayden: extra line (nit)

kimtoomey:
@sarahyablok sorry - new to this. Has this been pushed to production? Just following up w/ these guys.

sarahyablok:
@kimtoomey Yes this has been deployed in production.

kimtoomey:
@sarahyablok feedback from Regit on this: 
I'm struggling to replicate it during a test campaign or a in a draft environment.
Unrelated issue: In terms of the wrong CSS of some of the fonts I spoke about yesterday, here is an example beneath with one of the campaigns where text is the wrong colour using a custom style.

regit

PR #226, opened 2017-06-29
Add metaContains displayCondition

sarahyablok:
Example:

```
var modal = new pathfora.Message({
  id: 'meta-contains-modal',
  layout: 'modal',
  msg: 'This is a modal',
  displayConditions:{
    metaContains: [
      {
        property: 'og:type',
        content: 'website'
      }
    ]
  }
});
```

matches: 

```
<meta property="og:type" content="website" />
```


test/pathforaSpec.js
```diff
@@ -3899,6 +3899,53 @@ describe('Widgets', function () {
...
+    it('should not show if the meta rules are not met', function () {
+      pathfora.initializeWidgets([modal]);
+      expect($('#' + id).length).toBe(0);
+    });
+
+    it('should not show if a single meta rule is met', function () {
+      $('head').append('<meta property="og:type" content="product">');
+      $('head').append('<meta name="description" content="cool description here">');
+      pathfora.initializeWidgets([modal]);
+      expect($('#' + id).length).toBe(1);
```
markhayden: shouldn't this be 0 if we dont want it to be shown if both rules are not met?

test/pathforaSpec.js
```diff
@@ -3899,6 +3899,53 @@ describe('Widgets', function () {
...
+      $('meta').remove();
+    });
+
+    it('should not show if both rules are met', function () {
```
markhayden: think this should say "should show if..." since we are validating that it does in fact show.

test/pathforaSpec.js
sarahyablok: Whoooops! we definitely do want it to show if both at met. The description of the test is wrong.

test/pathforaSpec.js
sarahyablok: I did it again! 😱

test/pathforaSpec.js
markhayden: in the case of this one dont think its just a typo. the module will show if 1 of 2 conditions are met...which is why this test is passing :) so might want to look at that piece. just verified on a local test also.

src/rollup/display-conditions/meta-checker.js
```diff
@@ -0,0 +1,46 @@
...
+        if (rule.hasOwnProperty(key)) {
+          var val = meta[i].getAttribute(key);
+
+          if (!val || val !== rule[key]) {
+            phraseValid = false;
+            break;
+          } else {
+            phraseValid = true;
+          }
+        }
```
markhayden: just a hunch, have not tested. but we probablyi need a `phraseValid = false` here and a continue after the `phraseValid = true` on line 30 to ensure its not being shown when one of two conditions is met....i think

PR #227, opened 2017-08-04
Fix pageview display condition bug and update dependancies.

sarahyablok:


jstrubinger:
hi @markhayden ! Looks like the Personalize campaigns are now working? I tested in Investor's Business Daily Personalize campaign and their preview is showing an error

error

araddon:
just spoke with julie:  This is happening because of a cloudflare page redirect rule (i assume) not liking this url (you can test to verify)
* does not work:   http://www.investors.com/?test 
* does work:  http://www.investors.com?test

So, it works on urls other than home.  We could potentially not add the trailing slash on the preview from webadmin?

PR #247, opened 2017-10-11
Remove error for missing seerid

### LINKED ISSUE #241: Error when missing seerid cookie
seerid

To replicate:
Visit url in new incognito window, or clear cookies: https://lytics.github.io/pathforadocs/examples/preview/layouts/modal/contentRecommend.html

sarahyablok:
closes #241

Extra tiny fix to ensure we're not erroring on the first page view when a default content object is defined.


src/rollup/recommendations/recommend-content.js
```diff
@@ -55,7 +55,7 @@ export default function recommendContent (accountId, params, id, callback) {
   var seerId = readCookie('seerid');
 
   if (!seerId) {
-    throw new Error('Cannot find SEERID cookie');
+    callback([]);
```
markhayden: this is fine but do we need to ensure that callback is a function here and above ( i realize the above is already in production ). ever a chance of not having a callback function here?

sarahyablok: Should probably never happen, but I'll add a type check in case of user error.

markhayden: better safe than sorry :)

PR #256, opened 2017-11-10
Break up tests into multiple relevant files

### LINKED ISSUE #255: Break up tests into more meaningful files
Currently, we have a giant `pathfora.spec.js` that houses all our tests. The goal of this issue is to break it into multiple relevant files that are easier to maintain.

prosdev:
This PR closes #255. Tests from the giant `pathfora.spec.js` are split into the sections that reflect the features they're testing...mostly. We should create a ticket to clean up those individual files to make sure they're in the right place and leveraging Jasmine BDD principles, and do general code cleanup.

For now, the following was accomplished: 
- Split tests over multiple files. 
- Ensure no change in coverage. 

"files"


prosdev:
Thanks for looking over this, @sarahyablok!

PR #263, opened 2017-11-29
Add form data to confirmAction callback.

sarahyablok:
- Data returned from a user submitted form is now included in the parameters for the callback of `confirmAction`.
- We retain support for `onSubmit` callback, but note that it will become deprecated as `confirmAction` now can handle all cases `onSubmit` does.


src/rollup/widgets/actions/construct-widget-actions.js
```diff
@@ -209,48 +176,92 @@ export default function constructWidgetActions (widget, config) {
...
-              setTimeout(function () {
-                closeWidget(widget.id, true);
-              }, delay);
-            }
+      if (typeof widgetOnButtonClick === 'function') {
+        widgetOnButtonClick(event);
+      }
+
+      if (shouldClose) {
+        if (config.layout !== 'inline' && typeof config.success === 'undefined') {
```
prosdev: I find it useful to do `config.success == null` vs `typeof config.success === 'undefined'`, but I see that it's a pattern in this code base. Up to you.

src/rollup/widgets/actions/construct-widget-actions.js
```diff
@@ -209,48 +176,92 @@ export default function constructWidgetActions (widget, config) {
...
+          closeWidget(widget.id, true);
+          widgetOnModalClose(widget, config, event);
+
+        // show success state
+        } else {
+          addClass(widget, 'success');
+
+          // default to a three second delay if the user has not defined one
+          var delay = typeof config.success.delay !== 'undefined' ? config.success.delay * 1000 : 3000;
```
prosdev: Could do `config.success && config.success.delay` to check that success object

test/widget.spec.js
```diff
@@ -519,6 +519,36 @@ describe('Widgets', function () {
     expect(modal.confirmAction.callback).toHaveBeenCalled();
   });
 
+  it('should trigger callback function after pressing action with form data.', function () {
```
prosdev: Thanks for adding a test! Makes it much easier to try and read the code above.

src/rollup/widgets/actions/construct-widget-actions.js
sarahyablok: Adding the check for `config.success`. I am going to leave the typeof undefined check on delay since there is a chance it will be `0`.

PR #258, opened 2017-11-13
Async Error Handling for Pathfora Forms

gomezlavin:
This PR extends the Pathfora Form default behavior to now support asynchronous calls, which let the form wait for a callback that holds a success state or an error, and handles it accordingly.
In the case of a successful callback, the form behaves as normal, presenting the success modal to users.
In the case of an error, the form presents an error message north of the submit cta, which was previously defined in the Form's config object.

Here's an example of how to use it:
```<!DOCTYPE html>
<html>
  <head>
    <title>Custom button text</title>
  </head>
  <body>
    <h1>Custom button text example</h1>
    <script src="https://c.lytics.io/api/tag/0d47e6304d12858905b4f111090a68a1/lio.js"></script>
    <script src="../dist/pathfora.js"></script>
    <script>
      var modal = new pathfora.Form({
        id: 'mr-lytics-modal',
        className: 'mr-lytics-modal',
        layout: 'modal',
        branding: false,
        okMessage: 'Yes please',
        cancelMessage: 'No thanks',
        msg: 'Share your email for FREE SHIPPING on your first purchase.',
        success: {
          headline: "Free Shipping",
          msg: 'Your FREE SHIPPING on first purchase will be applied automatically at checkout.',
          delay: 0,
          okShow: true,
          okMessage: 'Continue Shopping',
          confirmAction: {
            name: 'cta_modal_success_continue_shopping',
            callback: {}
          }
        },

        formElements: [
          {
            'type': 'email',
            'name': 'email',
            'placeholder': 'Enter Email',
            'required': true
          },
          {
            'type': 'errorMessage',
            'name': 'errorMessage',
            'msg': 'Error message to be displayed to the user.'
          }
        ],

        displayConditions: {
          showDelay: 0
        },

        confirmAction: {
          waitForAsyncResponse: true,
          callback: function(id, obj) {

          },
          asyncCallback: function(id, obj, cb) {
            cb(true);
          }
        }
      });

      pathfora.initializeWidgets([modal])
    </script>
  </body>
</html>```

PR #265, opened 2017-12-04
Async Error Handling for Pathfora Forms: v2 (introduction of formStates field)

sarahyablok:
After merging #263 we needed to make some further updates to achieve the goal of #258, which was to include an handle an error state from an async call in a `confirmAction` callback. 

To do so we have deprecated the top level `success` field in favor of a field `formStates` which nests a `success` and `error` state.

We have added the `waitForAsyncResponse` field for controlling the addition of the success/error callback in the parameters of the `confirmAction` callback.

Example:

```js
    var module = new pathfora.Form({
      id: 'success-form-cbs',
      msg: 'subscription',
      headline: 'Header',
      layout: 'slideout',
      confirmAction: {
        waitForAsyncResponse: true,
        callback: function (name, payload, cb) {
          if (payload.data[0].value === "test") {
            cb(true); // calling the cb with "true" means the action was successful, and we should show the success state.
            return;
          }
          cb(false); // calling the cb with "false" means the action was not successful, and we should show the error state.
        }
      },
      formStates: {
        success: {
          headline: 'test',
          msg: 'a custom success message',
          delay: 0,
        },
        error: {
          headline: 'test',
          msg: 'a custom error message',
          delay: 0,
        }
      }
    });
    window.pathfora.initializeWidgets([ module ]);
```


src/rollup/data/tracking/track-widget-action.js
```diff
@@ -52,13 +52,23 @@ export default function trackWidgetAction (action, widget, htmlElement) {
     pathforaDataObject.cancelledActions.push(params);
     break;
   case 'success.confirm':
-    params['pf-widget-action'] = !!widget.success && !!widget.success.confirmAction && widget.success.confirmAction.name || 'success confirm';
+    params['pf-widget-action'] = !!widget.formStates.success && !!widget.formStates.success.confirmAction && widget.formStates.success.confirmAction.name || 'success confirm';
```
prosdev: I think this will throw an error, but can `formStates` come through as `null` or `undefined`? We may want to do a check here `widget.formStates && ...` and elsewhere

src/rollup/form/construct-form-state.js
```diff
@@ -0,0 +1,62 @@
...
+  case 'error':
+    obj = config.formStates.error;
+    defaultMsg = 'There was an error receiving with your submission.';
+    defaultHeadline = 'Error';
+    break;
+  default:
+    throw new Error('Unrecognized formState: ' + name);
+  }
+
+  var elem = document.createElement('div');
```
prosdev: It seems we do in some places in pathfora, but it would be nice to be consistent. Empty line after variable declarations, yes ... or no?

src/rollup/widgets/actions/construct-widget-actions.js
```diff
@@ -234,35 +247,31 @@ export default function constructWidgetActions (widget, config) {
...
+          // include the data from the form if we have it.
           if (data) {
             param.data = data;
           }
 
-          config.confirmAction.callback(callbackTypes.MODAL_CONFIRM, param);
-        }
-      }
+          // if waitForAsyncResponse we will handle the states as part of the callback
+          if (config.confirmAction.waitForAsyncResponse === true) {
```
prosdev: I see that the outer check `==='object'`. Will it be a problem if it's an empty `{}`?

test/widget.spec.js
```diff
@@ -744,11 +744,168 @@ describe('Widgets', function () {
...
+
+      email = form.find('input[name="email"]');
+      expect(email.length).toBe(1);
+      email.val('bad@example.com');
+      form.find('.pf-widget-ok').click();
+
+      success = widget.find('.success-state');
+      expect(success.length).toBe(1);
+      error = widget.find('.error-state');
+      expect(error.length).toBe(1);
```
prosdev: seems like a random line break

src/rollup/form/construct-form-state.js
sarahyablok: Yeah, I am super sorry this is inconsistent. I kind of have my own style on this, (which should not be the standard for pathfora, but just to explain) I usually don't have a line break if the next line I am making some change to the variable I just defined. This is dumb though, we should standardize it and enforce it with lint. (whether that's declaring all vars at the top of the function, or just enforcing a line break after all definitions). I will include this on the upcoming pathfora roadmap ticket.

src/rollup/widgets/actions/construct-widget-actions.js
sarahyablok: I don't think so. If `config.confirmAction` is an empty object, then `config.confirmAction.waitForAsyncResponse` is undefined. This is essentially equivalent to doing:

```js
var t;
if (t === true) {
  // do something
}
```

And in that case `t === true` just evaluates to false.

src/rollup/form/construct-form-state.js
prosdev: 👍

src/rollup/widgets/actions/construct-widget-actions.js
prosdev: Ahhh yes.

gulpfile.js
```diff
@@ -240,7 +240,7 @@ gulp.task('docs:watch', function () {
   });
 });
 
-gulp.task('docs:hbs', function () {
+gulp.task('docs:hbs', ['build:rollup'], function () {
```
prosdev: aha!

PR #279, opened 2018-02-01
added body to the custom form color variables

kristinh12:
This fixes a problem where the custom form labels on a modal have the default grey color, instead of the color of the msg text set in the UI.  

I added a body variable to set-custom-colors.js to get and set the color of the .pf-widget-body class, which includes the labels. 

I also added the default grey color for pf-widget-body to the pathfora CSS.

I tested it with the same custom fields and colors that my customer was trying to use:
screen shot 2018-02-01 at 10 31 01 am

Closes: https://github.com/lytics/pathforajs/issues/277


kristinh12:
@sarahyablok It looks like there's still files from the dist folder even after running those npm commands.

sarahyablok:
@kristinh12 Yeah, that's fine. By extra "stuff" I meant the code coverage items that get added to the js files in dist. But you're good now.

src/rollup/widgets/colors/set-custom-colors.js
```diff
@@ -27,7 +27,8 @@ export default function setCustomColors (widget, colors) {
       required = widget.querySelectorAll('.pf-required-flag'),
       requiredAsterisk = widget.querySelectorAll('span.required'),
       requiredInline = widget.querySelectorAll('[data-required=true]:not(.pf-has-label)'),
-      socialBtns = Array.prototype.slice.call(widget.querySelectorAll('.social-login-btn'));
+      socialBtns = Array.prototype.slice.call(widget.querySelectorAll('.social-login-btn')),
+      body = widget.querySelector('.pf-widget-body');
```
prosdev: Ahhhhh ok! This makes sense!

src/rollup/widgets/colors/set-custom-colors.js
```diff
@@ -144,4 +145,8 @@ export default function setCustomColors (widget, colors) {
       msg[i].style.setProperty('color', colors.text, 'important');
     }
   }
+
+  if (body && colors.text) {
+    body.style.setProperty('color', colors.text, 'important');
```
prosdev: 🎉

kristinh12: Thanks for helping!

Issue #275, opened 2017-12-27
remove watermark


markhayden:
lets make the watermark off by default...seems to be a nearly unanimous request.

PR #282, opened 2018-02-02
Remove default watermark

kristinh12:
Unless I'm missing something, this was a really simple fix.  Just changed branding to false in the globals default settings so that we're not making overrides for nearly all of our clients.

Closes: https://github.com/lytics/customerprojects/issues/270
Closes: https://github.com/lytics/customerprojects/issues/256
Closes: https://github.com/lytics/pathforajs/issues/275


kristinh12:
Looks like I need to update the test for the branding as well.

prosdev:
@kristinh12 Can you squash all the commits together -- It'd be nice to have the passing tests and fix on the same commit.

kristinh12:
I don't know that there are any, but yes, we'd need to.  @sarahyablok is there an easy way to see which clients are actually using the watermark?

sarahyablok:
@kristinh12 @prosdev I think the goal is to remove it. I don't believe there are any customers who want it or use it currently on their modals, so the goal is just to remove it completely.

PR #288, opened 2018-03-22
Accessibility Part 2: Limit focus tab cycle for modals & sitegate

sarahyablok:
For large, screen covering widgets such as modals and Site gates, we want to keep the tab cycle of focus within the widget elements, and prevent access from focusing on the page below.


test/pathfora.spec.js
```diff
@@ -12,6 +11,14 @@ var jstag = {
 pathfora.utils.saveCookie('seerid', 123);
 pathfora.enableGA = true;
 
+function createAndDispatchKeydown (key, target) {
```
prosdev: 👍

test/pathfora.spec.js
```diff
@@ -23,6 +30,36 @@ describe('Pathfora', function () {
     pathfora.clearAll();
   });
 
+  it('should keep focus during a tab cycle in a modal or site gate', function (done) {
```
prosdev: nice test! 🎉

Issue #287, opened 2018-03-15
Adding a Footer Feature


kristinh12:
We have had many clients request that text be added at the bottom of modals, often legal copy.  It would be really useful to have an option for a footer below the CTA.  

Presently our only way to accomplish this is to have clients add text to their background image which often gets cut off.

PR #293, opened 2018-04-11
Footer feature

kristinh12:
Closes: https://github.com/lytics/pathforajs/issues/287

-added a div with the class pf-widget-footer to modals and slideouts
-in construct-widget-layout.js I set the text to the footerText setting
-updated the docs to include this new setting (let me know if I should add it somewhere else)
-added css for pf-widget-footer in widgets-general.less and slideouts.less

Some images of the footer:
screen shot 2018-04-11 at 3 03 42 pm

screen shot 2018-04-11 at 4 09 00 pm


kristinh12:
Does the codecov/patch fail mean I need to add a test?

markhayden:
yah it means the coverage for the pr did not stay above the minimum threshold required..which i think is 50%. generally we want to have 100% coverage on all new updates but there are edge cases.

src/less/widgets/slideouts.less
```diff
@@ -32,6 +32,10 @@
     font-size: 20px;
   }
 
+  .pf-widget-footer {
```
markhayden: do we need to address the other types here as well (modal)?

src/rollup/widgets/construct-widget-layout.js
```diff
@@ -333,4 +334,8 @@ export default function constructWidgetLayout (widget, config) {
   if (config.msg) {
     widgetMessage.innerHTML = config.msg;
   }
+
+  if (config.footerText) {
+    widgetFooter.innerHTML = config.footerText;
```
markhayden: should be able to add lots of tests for this. one to ensure the div gets populated properly when there is text to drop in and another to ensure that it does not get added at all for widgets with no `footerText`. both on every type of widget it can effect. could also test empty string vs missing undefined key if you want to go wild.

src/less/widgets/widgets-general.less
```diff
@@ -38,6 +38,12 @@
     color: @text;
   }
 
+
+  .pf-widget-footer {
```
markhayden: curious if this should be `display:none` by default as it mixed with the slideout style above might effect widgets in production since they'd be missing the necessary override of the `pf-widget-footer`. might be worth testing against some variations in production to make sure?

src/less/widgets/slideouts.less
```diff
@@ -32,6 +32,10 @@
...
+    margin-top: 10px;
```
markhayden: see below

kristinh12: @markhayden I already checked that nothing is visually different about the modals/slideouts if there is no footerText setting.  I had it as padding-top at first which did change the appearance of the modal since it created a div with a height of 10, but margin didn't have that effect.

markhayden: perfect, just making sure

sarahyablok:
forgot to mention, but you should run `npm run prod` before committing. That will remove all the extra stuff that makes the diff look huge.

src/less/widgets/widgets-general.less
```diff
@@ -38,6 +38,12 @@
...
+    font-size: 12px;
+    text-align: left;
```
sarahyablok: Idk if this would be a hassle, but I noticed on all widgets this is left aligned. Some widgets like modal forms have center aligned text by default. Would be great to match the text alignment with the text above.

kristinh12:
Ok, this could be reviewed again now. @markhayden and @sarahyablok 

screen shot 2018-04-12 at 11 48 35 am

screen shot 2018-04-12 at 1 35 38 pm

screen shot 2018-04-12 at 1 35 54 pm

kristinh12:
I did run `npm run prod` this time.

Issue #298, opened 2018-06-04
Recommend content campaign will not preview if document has no meta description.


shindakun:
Issue: If the content has a blank description modal will not display.

![image](https://user-images.githubusercontent.com/319128/40933239-b8bf391c-67e5-11e8-8786-89e263c38bb1.png)
![image](https://user-images.githubusercontent.com/319128/40933259-c454e29a-67e5-11e8-86f9-f1cac3af55c4.png)

Expected: Content should display even if the description field happens to be blank.

kristinh12:
Fix for this was merged awhile back, going to close.

PR #300, opened 2018-06-27
Recommendation description bug

kristinh12:
Issue: https://github.com/lytics/pathforajs/issues/298

Previously if a recommendation was missing a description, the modal wouldn't display.  

I added to the conditional to check for html tags before trying to set the color, and wrote a test.

Modal successfully displays:
screen shot 2018-06-27 at 2 11 19 pm


test.html
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE html>
```
prosdev: @kristinh12 Can we be a bit more descriptive with this file name? `test.html` doesn't give that much context. Maybe `widget-no-description-test.html` or something like that.

kristinh12: Ah I can probably just delete that, I wrote an actual test widget.spec.js.  The test.html was to try to duplicate the problem and have a visual of it working.

prosdev: I think these are super help though. Maybe we can add a folder to keep all of these? What do you think @sarahyablok ?

sarahyablok: We had an examples folder which got deleted cause it had bad examples. Perhaps for now we can delete this and file a ticket to create a new ticket for a new examples folder.

dist/pathfora.js
```diff
@@ -2581,11 +2581,11 @@ function setCustomColors (widget, colors) {
       contentUnit.style.setProperty('background-color', colors.actionBackground, 'important');
     }
 
-    if (colors.actionText) {
+    if (colors.actionText && contentUnitMeta.querySelector('h4')) {
```
sarahyablok: This is a tiny efficiency nit, but you can actually define this as a variable up in the jumble of vars at the top of this function, that saves us from doing the querySelector lookup twice.

```
var contentUnitMetaTitle = contentUnitMeta.querySelector('h4');

...

if (colors.actionText && contentUnitMetaTitle) {
  contentUnitMetaTitle.style.setProperty('color', colors.actionText, 'important');
}
```

test.html
prosdev: Sounds good! I'm ok with that! 👍

test/widget.spec.js
```diff
@@ -492,6 +492,38 @@ describe('Widgets', function () {
...
+        title: 'test',
+        default: true
+      }]
+    });
+    window.pathfora.initializeWidgets([modal]);
+
+    var widget = $('#' + modal.id);
+    expect(widget).toBeDefined();
+  });
+
```
sarahyablok: I believe there is a `.spec` file for recommendation widgets, perhaps this test is better suited there.

Issue #305, opened 2018-08-08
Add div around headline & body text (excluding image)


sarahyablok:
Currently, pathfora does not have a wrapper div around the headline text and the main content of the modal, that does not include the image. To allow for proper image floats (without setting a width on the image) we'd want a div that could be flexible:

For example:

<img width="852" alt="screen shot 2018-08-08 at 10 45 56 am" src="https://user-images.githubusercontent.com/5973633/43854698-8e491d2e-9af8-11e8-8838-f4acc358eb53.png">

This looks ok, but when we try to do this with a smaller content space the image is constrained to  just the text and button space, and with a width set, it looks odd.
<img width="872" alt="screen shot 2018-08-08 at 10 46 15 am" src="https://user-images.githubusercontent.com/5973633/43854710-9593ccbe-9af8-11e8-97f2-427e35095a5d.png">

kristinh12:
@sarahyablok should we add this for both modals and slideouts?

sarahyablok:
@kristinh12 I would guess all module types.

kristinh12:
https://github.com/lytics/pathforajs/pull/317

PR #315, opened 2018-11-07
Div for image

kristinh12:
Closes: https://github.com/lytics/pathforajs/issues/305

I added a div in each inline and modal template with the class name "pf-widget-text".  Open to suggestions, but all the names that seem logical to me have already been taken.

Slideouts and bars already had a structure where the headline and message could be in a separate div than the image so I didn't add pf-widget-text for them.

I then edited the construct-widget-layout so that depending on which type of layout, the image will be added to a different div so that we are able to float the text to the left or right and have the image on the other side.  This will help us meet a lot of requests for different image placements from clients and hopefully create templates for common requests that @sarahyablok  and I worked on during the process hackday.

Some images of different widgets and their dom structure:

screen shot 2018-11-07 at 1 22 39 pm
screen shot 2018-11-07 at 1 23 11 pm

screen shot 2018-11-07 at 1 24 13 pm
screen shot 2018-11-07 at 1 24 08 pm

screen shot 2018-11-07 at 1 24 36 pm
screen shot 2018-11-07 at 1 25 00 pm


src/rollup/widgets/construct-widget-layout.js
```diff
@@ -187,12 +187,14 @@ export default function constructWidgetLayout (widget, config) {
 
   // NOTE Set the image
   if (config.image) {
+    var widgetImage = document.createElement('img');
+    widgetImage.src = config.image;
+    widgetImage.className = 'pf-widget-img';
     if (config.layout === 'button') {
```
prosdev: Should rework this logic. It looks like we can clean it up so that there isn't two `widgetContent.appendChild(widgetImage);`.

kristinh12: Where are you seeing two of them?

test.html
```diff
@@ -10,29 +10,11 @@ <h1>Action widget example</h1>
...
-        recommend: {
-          collection: 'owifnwkqlbf'
-        },
-        content: [{
-          url: 'https://www.getlytics.com/blog/post/know_your_data',
-          title: 'test',
-          description: 'I am a sample test description.',
-          default: true
-        }]
+        variant: 2,
```
prosdev: Add tests for the `inline` and `modal` templates with images?

src/rollup/widgets/construct-widget-layout.js
prosdev: Ignore. I misread `widgetContent` and `widgetBody`

kristinh12:
@prosdev the test.html file was just for testing appearance as I went along, but I just added several tests to the test coverage for the changes I made.

sarahyablok:
My concern with this PR is, will it break any existing CSS that folks have written? I suspect not unless people are abusing the `>` operator, but we should text and review with some known active CSS files.

kristinh12:
Ok, I'll try to find some recent modals we've designed that I could borrow the CSS for.

PR #317, opened 2018-11-26
pf-widget-text div for image flexibility

kristinh12:
Most of these changes were already approved previously, but this PR does the following:

1.) Adds a pf-widget-text for inline and modal layouts that surrounds the headline and form or message.
2.) Edits the code slightly for placing an image so that image placement can be more flexible, such as float left or right rather than centered at the top.
3.) Added tests for this new div
4.) updated the css customization docs

I also checked several client modals to make sure this change would not affect them.


kristinh12:
https://github.com/lytics/pathforajs/pull/317

PR #322, opened 2019-01-21
Account for multiple ga instances

### LINKED ISSUE #321: GA tracking code is not properly setting instance name on send
when sending events to google analytics we are just sending a general `ga('send'...)` request. that works unless the instance of GA is named. at some point GTM defaulted to naming their instances which broke the default Pathfora event tracking. The following snippet is from our core tag as an example of how we pull the tracking ids and append names. 

```
var trackers = ga.getAll(),
    trackingIds = [];

for (var i = 0, len = trackers.length; i < len; i++) {
    var id = trackers[i].get('trackingId'),
        name = trackers[i].get('name');

    id && trackingIds.push({"name":name, "id":id});
}

for (var j = 0; j < trackingIds.length; j++) {
    var name = trackingIds[j].name;
    window.ga(name + '.send', 'event', 'somename', 'somevalue', {'nonInteraction': 1});
}
```

sarahyablok:
closes #321 

This fixes a bug where events were not being collected in GA if multiple instances were set up (usually through gtm).


dist/pathfora.js
```diff
@@ -1021,25 +1021,30 @@
...
-        'send',
-        'event',
-        'Lytics',
-        data['pf-widget-id'] + ' : ' + gaLabel,
-        '',
-        {
-          nonInteraction: true
-        }
-      );
+      for (i = 0; i < trackers.length; i++) {
```
prosdev: Since we aren't using `i` outside of this scope, I would bring it here. 
```suggestion
      for (var i = 0; i < trackers.length; i++) {
```

src/rollup/data/request/report-data.js
```diff
@@ -9,26 +9,31 @@ import window from '../../dom/window';
...
-      'send',
-      'event',
-      'Lytics',
-      data['pf-widget-id'] + ' : ' + gaLabel,
-      '',
-      {
-        nonInteraction: true
-      }
-    );
+    for (i = 0; i < trackers.length; i++) {
```
prosdev: Since we aren't using i outside of this scope, I would bring it here.

```suggestion
    for (var i = 0; i < trackers.length; i++) {
```

sarahyablok:
:rage: Github suggest trolled me

PR #326, opened 2019-04-12
Inline and Modal background image layouts

### LINKED ISSUE #304: Add a background image option for modals
One of the most requested CSS  changes is adding a background image.

We would like to be able to have this as an option from the UI eventually, and create a template with a background image positioned at the top of the modal.

For now, it can be added as a pathfora feature that would require an override.

Eventual goal would look something like this:
image

krothenbaum:
Closes #304 

Adds a new layout variant with a background image for modal and inline layouts. Includes option for background to cover the entire widget or positioned to cover the top, bottom, right or left half of the widget.

Added tests and updated documentation.

~~**Note:** I'm not sure about the giant line change on the `pathfora.js` file. I'm assuming a build command I ran modified it. What's the best way to revert that change?~~ Ran `npm run prod`.

Modal
image

Inline
image


docs/docs/examples/src/layouts/inline/backgroundImage.js
```diff
@@ -0,0 +1,13 @@
+var module = new pathfora.Subscription({
+  id: 'message-inline',
+  layout: 'inline',
+  position: '.inline-module',
+  headline: 'Welcome!',
+  msg: 'Please sign up for our newsletter for more updates.',
+  variant: 4,
```
sarahyablok: Might be something we need to discuss in person, but what was the deciding factor on making this a separate variant? Given the existing patterns with variants, it does sort of make sense. But in general I think we want to avoid using variants in the future as they are kind of a bad API design, and I think we'd like to support background images on other variant modals such as content recommendations.

docs/docs/layouts/inline.md
```diff
@@ -85,4 +89,50 @@ Define the featured image you would like to use for the module.
 
 ![Image Inline Module](../examples/img/layouts/inline/image.png)
 
-<pre data-src="../../examples/src/layouts/inline/image.js"></pre>
\ No newline at end of file
+<pre data-src="../../examples/src/layouts/inline/image.js"></pre>
+
+## background image
```
sarahyablok: This should be `backgroundImage` the titles are formatted as they appear in the config.

docs/docs/layouts/inline.md
```diff
@@ -85,4 +89,50 @@ Define the featured image you would like to use for the module.
...
+
+Define the background image and position you would like to use for the module.
+
+**Note:** This setting is only valid for modules with a variant value of 2.
```
sarahyablok: value of 4 (although I vote that this is subject to change).

docs/docs/layouts/modal.md
```diff
@@ -56,3 +60,57 @@ Define the featured image you would like to use for the module.
 ![Image Slideout Module](../examples/img/layouts/modal/image.png)
 
 <pre data-src="../../examples/src/layouts/modal/image.js"></pre>
+
+## background image
```
sarahyablok: same on title formatting

docs/docs/layouts/modal.md
```diff
@@ -27,6 +27,10 @@ Variant determines any extra content (dictated by extra keys in the config) that
     <td>3</td>
     <td>module includes a <a href="../../content_recommend">content recommendation</a></td>
   </tr>
+  <tr>
+    <td>4</td>
+    <td>module icludes a background image</td>
```
sarahyablok: `includes`

test/acceptance/widget.spec.js
```diff
@@ -397,6 +397,70 @@ describe('Widgets', function () {
...
+      variant: 4,
+      backgroundImage: {
+        src: '../docs/docs/assets/desk.png'
+      }
+    });
+    pathfora.initializeWidgets([modal, inline]);
+     var modalWidget = $('#' + modal.id),
+        inlineWidget = $('#' + inline.id);
+    expect(modalWidget.find('.pf-widget-background-image')).toBeDefined();
+    expect(inlineWidget.find('.pf-widget-background-image')).toBeDefined();
```
sarahyablok: I would go ahead and check the inline styling here too.

test/acceptance/widget.spec.js
```diff
@@ -397,6 +397,70 @@ describe('Widgets', function () {
     expect(inlineWidget.find('.pf-widget-content').find('img').html()).toBeDefined();
     expect(modalWidget.find('.pf-widget-text').find('img').html()).toBeUndefined();
   });
+  //variant 4
+  it('should append div with pf-widget-background-image for modal and inline layout', function() {
```
sarahyablok:
```suggestion
  it('should support backgroundImage src', function() {
```

test/acceptance/widget.spec.js
```diff
@@ -397,6 +397,70 @@ describe('Widgets', function () {
...
+      }
+    });
+    pathfora.initializeWidgets([modal, inline]);
+     var modalWidget = $('#' + modal.id),
+        inlineWidget = $('#' + inline.id);
+    expect(modalWidget.find('.pf-widget-background-image')).toBeDefined();
+    expect(inlineWidget.find('.pf-widget-background-image')).toBeDefined();
+  });
+
+  it('should append position class to pf-widget-background-image and pf-widget-text for modal and inline layouts', function() {
```
sarahyablok:
```suggestion
  it('should support backgroundImage position', function() {
```

sarahyablok: My general thought is, these test names don't need to be super descriptive. Just defining what feature it is attempting to test is enough. Also using the config setting name is more informative to other devs than the class names.

PR #328, opened 2019-04-26
Update toggle.md

shindakun:
no related issue
old link 404'd

@sarahyablok


shindakun:
Sounds like this will be addressed via redirects. Closing.

sarahyablok:
We should still update the link. ;)

PR #330, opened 2019-06-25
Fix bug payload of callback for custom forms with checkboxes

sarahyablok:
Found a bug while developing new training content.

There was a bug with the data passed to the `confirmAction` callback of widgets with a custom form that used checkboxes or radio buttons. It would add the value of all possible choices in the payload, not just the items that were selected. This does not affect tracking data that was sent to Lytics. Only the payload of a custom callback.

I also then realized somehow the linter stopped picking up the test files. Fixed that and any lingering lint issues that happened after that broke, and decided to update our npm dependencies.


test/acceptance/widget.spec.js
```diff
@@ -653,6 +653,65 @@ describe('Widgets', function () {
...
+    });
+
+    pathfora.initializeWidgets([modal]);
+
+    var widget = $('#' + modal.id);
+    widget.find('input[name="email"]').val('test@example.com');
+    widget.find('input[name="subscription_feeds"]')[2].checked = true;
+    spyOn(modal.confirmAction, 'callback');
+    expect(modal.confirmAction.callback).not.toHaveBeenCalled();
+  widget.find('.pf-widget-ok').click();
```
prosdev: need an indentation here.

PR #335, opened 2019-08-23
remove static lytics cookie name and handle identity overrides

markhayden:
When requesting content recommendations the the `seerid` and `_uids` respectively were hard coded. They Lytics JavaScript tag allows you to customize both the default cookie as well as the key/value used for identification. This PR ensures that both of those scenarios are properly accounted for in order to get the most accurate content recommendation.


src/rollup/recommendations/recommend-content.js
```diff
@@ -59,17 +59,46 @@ export default function recommendContent (accountId, params, id, callback) {
...
+
+  // check for custom cookie name in jstag config
+  if (window.jstag && window.jstag.config && window.jstag.config.cookie !== '') {
+    storedCookieName = window.jstag.config.cookie;
+  }
+
+  // attempt to get value from stored cookie
+  userByFieldValue = readCookie(storedCookieName);
+
+  // override everything is key/value have been explicitly set for user
```
prosdev:
```suggestion
  // override everything if key/value have been explicitly set for user
```

PR #342, opened 2019-10-21
Tear down event listeners on clearall

### LINKED ISSUE #341: clearAll() function does not properly break down existing eventListeners
in the case of window scroll listeners, the `clearAll()` function is not blowing them away properly resulting in potential for showing the same widget over and over in the case of a single page app. in order to fix we should iterate and kill all pending event listeners related to pathfora as well as pass through the remaining display conditions to ensure there are no other listeners or queues of any sort that might be an issue down the road. 

https://github.com/lytics/pathforajs/blob/develop/src/rollup/display-conditions/scroll/init-scroll-watchers.js#L16

cbrock:
Closes #341


.gitignore
```diff
@@ -6,6 +6,5 @@ docs/site/
 docs/docs/examples/preview
 .DS_Store
 .env.json
-package-lock.json
 coverage/
-yarn.lock
\ No newline at end of file
+yarn.lock
```
humanchimp: why did you choose package-lock.json rather than yarn.lock?

cbrock: The current setup steps use `npm ...` so `package-lock.json` is created by default.

humanchimp: i'm concerned about preferring npm on some projects and yarn on others. In my opinion, we should use yarn everywhere. Doesn't need to happen here (although this PR is introducing +10000 new lines due to its inclusion of package-lock.json)

cbrock: I think it's reasonable to standardize on `yarn`. Please keep in mind this is a WIP/draft PR as I get up to speed on the repo 😄

humanchimp: sorry for premature feedback

cbrock: Updated to prefer yarn over npm.

PR #299, opened 2018-06-26
Add Priority Options Param

sarahyablok:
For orchestrate, we will be initializing the "candidate" widgets in in an ordered priority. This PR adds support for a third parameter to the `initializeWidgets` method.

```
pathfora.initializeWidgets([messageBar, modal], null, {priority: 'ordered'});
```

the options param will recognize a key `priority` which, if set to `ordered` will only initialize the first valid widget in the array initialized (according to display conditions, etc).

As part of this, we have deprecated the old `priority` display condition.

This PR also introduced a few structure changes to the repo:
- New folder `src/rollup/validation` folder now exists to hold our validation functions.
- We now have unit and acceptance tests. The unit tests will run through rollup to import single functions to test from the src.
- lio and recommendations are now preloaded up front when necessary.
- New folder `src/rollup/widgets/has` contains simple "has" (boolean) functions that are repeatable across the code base.
- `initializeWidgetArray` refactored to be recursive to support breaking of the loop even in async callbacks.

PR #350, opened 2019-10-24
persist settings in localStorage instead of document.cookie

### LINKED ISSUE #344: add api for performing CRUD operations against localStorage

### LINKED ISSUE #345: define JSON model for storing interaction data in localStorage

### LINKED ISSUE #346: expiration manager against localStorage payload
- cull expired data
- extend active date: on `.getItem`, update the expiration date

### LINKED ISSUE #347: existing cookie migration to localStorage handler

### LINKED ISSUE #348: integrate new localStorage handler as core for all dependent widgets

humanchimp:
Closes #344 
Closes #345
Closes #346 
Closes #347
Closes #348


humanchimp:
I was thinking that we could abbreviate the keys we're using in localStorage.

`payload` => `$` and `expiresOn` => `@`.

humanchimp:
> I was thinking that we could abbreviate the keys we're using in localStorage.

done in ad7d242

src/rollup/utils/persist/read.js
```diff
@@ -0,0 +1,18 @@
...
+
+export default function read (key) {
+  var item = expiringLocalStorage.getItem(key);
+
+  if (item == null) {
+    item = readCookie(key);
+
+    if (item != null) {
+      deleteCookie(key);
+      expiringLocalStorage.setItem(key, item);
```
humanchimp: i suppose this should get set unconditionally, so the expiration date gets extended on read

test/acceptance/utils.spec.js
```diff
@@ -87,44 +87,19 @@ describe('Utils', function () {
...
-      };
-
-      setCookie('PathforaClosed_1', '1%257C1502732355490'); // double encoded
       setCookie('PathforaImpressions_2', '1|293847239874932871');
       sessionStorage.setItem('PathforaRecommend_2', '{"somejson": "here"}');
       setCookie('PathforaImpressions_3', '%badval%');
 
       pathfora.utils.updateLegacyCookies();
 
-      expect(getCookie('PathforaClosed_1')).toEqual('1%7C1502732355490');
```
markhayden: any reason, in particular, we removed the additional tests? overkill i assume?

src/rollup/utils/persist/read.js
markhayden: definitely

src/rollup/utils/persist/expiring-local-storage.js
```diff
@@ -0,0 +1,53 @@
...
+      }
+      return value[PAYLOAD_KEY] || value;
+    }
+    return localStorage.getItem(key);
+  },
+
+  setItem: function (key, payload, expiresOn) {
+    if (!expiresOn) {
+      expiresOn = new Date();
+      expiresOn.setDate(expiresOn.getDate() + 365);
```
markhayden: do we want to expire in a year or 30 days? thought we talked about being more aggressive on the window but perhaps its a smaller problem when in localStorage. probably ok to leave as is for now so ist 1 to 1 with whats there then can change in a follow up PR if we so choose.

src/rollup/utils/persist/expiring-local-storage.js
```diff
@@ -0,0 +1,53 @@
...
+  try {
+    return JSON.parse(json);
+  } catch (e) {
+    // recover
+  }
+}
+
+export default {
+  getItem: function (key) {
+    var value = safeJsonParse(localStorage.getItem(key));
```
markhayden: do we need to ensure `localStorage` exists during init or something and log an error if it doesn't? should be an extreme edge case that we talked about but we still want to prevent errors in the event localStorage doesn't exist (possible i missed the check also)

src/rollup/utils/persist/expiring-local-storage.js
```diff
@@ -0,0 +1,53 @@
...
+
+    if (value && value[EXPIRES_KEY]) {
+      if (Date.parse(value[EXPIRES_KEY]) < Date.now()) {
```
markhayden: correct me if im wrong but it looks like this only gets ran when doing a read. which means the key (widget) is being evaluated by pathora, in which case we want to extend the expiration date. this looks like it will actually result in that stored value from being removed from localstorage.

e.g. i've closed a modal 30 days ago. i go back to the site 30days and 1 hour later but still initialie the widget. it triggers a read and as a result wipes that expired value when it should probably actually extend the expiration of that value. 

this likely brings up a couple questions / concerns with how we are handling expirations in general so it might be worth chatting and mapping that out so we are all on the same page and have something documented. 

might be the case that we need a flush method of sorts that looks at everything in local storage and expires them when they haven't been initialized. that flush method gets called on pathfora init. it would still result in the above example being wiped but it wouldn't be associated with the read....if that makes sense.

markhayden: one more thing we need to think about is how this reacts when the `priority` https://github.com/lytics/pathforajs/pull/299 option is passed. need to confirm this but i'd imagine in this case it stops evaluating the initialized widgets once one can be shown. this would also probably have the side effect of the expiration dates not being extended properly for the other widgets?

markhayden: defaulting to one year basically eliminates all of the above edge cases so maybe we focus on that for MVP and then regroup on how it should function in a perfect world?

test/acceptance/utils.spec.js
humanchimp: some of those assertions were concerned with double-encoding. At one point, there was a bug which was causing cookie values to be double-encoded, and then we had code for double-unencoding them to deal with cookies in the wild. Enough time has passed now (> the cookie expiration period) that we don't need to worry about cookies that were double-encoded, so we were able to remove the code and the tests. cc @sarahyablok 

More generally, keys and values stored in `localStorage` don't need to be URL encoded—localStorage manages to keep track of the content boundaries itself without cooperation from the API consumer. 😂

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: i can feature detect for `localStorage`. During planning, we decided it will be present in the browsers we care about (other than Opera Mini??? but I have never heard of anyone using it) But if `localStorage` is not present, what should we do? Nothing at all? Just exit with an error? That's approximately what will happen if `localStorage` is not present.

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: i just went with 1 year because that's what it was with cookies. We talked about extending the configuration to make this configurable

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: pretty sure this is not different from https://github.com/lytics/pathforajs/pull/350/files/f5300141314955d7db6fd95bae0724cb07e7b735#r338871395

We should be extending the expiration date on read

src/rollup/utils/persist/erase.js
```diff
@@ -0,0 +1,7 @@
+import expiringLocalStorage from './expiring-local-storage';
+import deleteCookie from '../cookies/delete-cookie';
+
+export default function erase (key) {
+  expiringLocalStorage.removeItem(key);
+  deleteCookie(key);
```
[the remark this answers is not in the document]
humanchimp: let's open an issue and remove the rest of the cookie-handling code next year?

src/rollup/utils/persist/read.js
humanchimp: done... just, not here. the change was made to the interface `expiringLocalStorage.getItem`. I added a test case for this...

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: Updated

src/rollup/utils/persist/erase.js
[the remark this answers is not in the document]
markhayden: makes sense. i'll open a ticket.

test/acceptance/utils.spec.js
markhayden: :+1:

src/rollup/utils/persist/expiring-local-storage.js
markhayden: yah that is kind of what i was thinking "The Pathfora Personalization SDK requires local storage. Unable to initialize." or something along those lines.

src/rollup/utils/persist/expiring-local-storage.js
markhayden: works for me. i think having this at a year also puts much less emphasis on some of the weirdness around extending expiration dates as well (given we have a year to solve it and at that point its unlikely that the interaction from a year previous is relevant anyway).

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: we are, elsewhere in pathfora, using `sessionStorage` which was introduced alongside `localStorage`... I guess we should have exit early if both are not defined, and claim we depend on the Web Storage API.

markhayden: /me nods

src/rollup/pathfora.js
```diff
@@ -58,6 +58,11 @@ import initializeInline from './inline/init-inline';
  * @class {function} Pathfora
  */
 var Pathfora = function () {
+  // feature detections
+  if (!('localStorage' in window) || !('sessionStorage' in window)) {
+    throw new Error('The Pathfora SDK requires the Web Storage API!');
+  }
```
humanchimp: i'm gonna write a test case for this

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: Done

src/rollup/utils/persist/expiring-local-storage.js
sarahyablok: :thinking: In our conversation, I thought we landed on 30 days? Also I am less concerned about it, cause every time we read it it gets extended by 30 days, right?

src/rollup/utils/persist/expiring-local-storage.js
```diff
@@ -0,0 +1,58 @@
...
+
+  setItem: function (key, payload, expiresOn) {
+    if (!expiresOn) {
+      expiresOn = new Date();
+      expiresOn.setDate(expiresOn.getDate() + 365);
+    }
+
+    var record = {};
+
+    record[PAYLOAD_KEY] = '' + payload;
```
sarahyablok: why do we do `'' + payload` here? Just to ensure it's a string?

src/rollup/utils/persist/expiring-local-storage.js
```diff
@@ -0,0 +1,58 @@
...
+    record[EXPIRES_KEY] = expiresOn;
+
+    localStorage.setItem(key, JSON.stringify(record));
+  },
+
+  removeItem: function (key) {
+    localStorage.removeItem(key);
+  },
+
+  ttl: function (key, payload, milliseconds) {
```
sarahyablok: Where do we use this method?

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: yeah, this casts as a string to match the behavior of `localStorage`. We _could_ store any serializable value instead, but that's not what localStorage does 🤷‍♂

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: we don't currently use it. i left it here because it was already implemented (some of this code was harvested from webadmin), and I am planning on using it for implementing the planned config item (ttl: '1M' or whatever)... but let's remove it from this PR.

humanchimp:
> 🤔 In our conversation, I thought we landed on 30 days? Also I am less concerned about it, cause every time we read it it gets extended by 30 days, right?

(For some reason i cannot reply inline to this comment)

I copied the expriration logic from what we were doing with cookies, because i was planning on implementing a config option, (`ttl: '1M'` or whatever), but we decided to hold off on the config item. I'm fine with making this 30d

humanchimp:
Notes from IRL discussion
- [x] cookie expiration date remains 1 year out
- [ ] eagerly migrate cookies on init
- [ ] eagerly cull expired `localStorage` records on init
- [x] do not eagerly extend expiration dates on records in `localStorage`

cc @markhayden @sarahyablok

humanchimp:
> Right now the only way it gets converted is on read, yes?

This turned out not the be the case. We eagerly migrate existing cookies [here](https://github.com/lytics/pathforajs/blob/02be42b19dfd12803830e735f65f68328183e23f/src/rollup/utils/cookies/update-legacy-cookies.js#L69)

humanchimp:
> Where do we use this method?

> ...but let's remove it from this PR.

We're using `.ttl` in tests, and rather than rewriting the tests, we decided to leave it in

src/rollup/utils/persist/expiring-local-storage.js
humanchimp: we decided to keep it, since it's used in tests and we plan to use it when we parameterize expiration

test.html
```diff
@@ -8,15 +8,10 @@ <h1>Action widget example</h1>
...
-        msg: 'Welcome to our website',
-        variant: 2,
-        image: 'https://lytics.github.io/pathforadocs/assets/lion.jpg',
-        headline: 'Headline'
-      });
-      window.pathfora.initializeWidgets([modal]);
+        pathfora.utils.store.ttl('current', 'test', 1000);
+        pathfora.utils.store.ttl('expired', 'test', -1000);
+
+
```
humanchimp: this change was not intended

PR #372, opened 2019-12-06
Exit Intent Refactor to Watcher System

### LINKED ISSUE #367: Exit Intent not working with Scroll Percentage Display Condition
When both display conditions are present for a modal, the scroll position takes precedence over the exit intent condition and the modal appears without exit intent.

sarahyablok:
closes #367 

- Refactor the Exit Intent display conditions code to use the "watcher" system used by other similar display conditions.
- Ensure that we're checking all previous watcher conditions in `validateWatchers`


src/rollup/display-conditions/exit-intent/register-exit-intent-watcher.js
```diff
@@ -0,0 +1,40 @@
...
+ * @exports registerExitIntentWatcher
+ * @params {string} selector
+ * @params {object} widget
+ * @returns {object} watcher
+ */
+export default function registerExitIntentWatcher () {
+  var watcher = {
+    positions: [],
+    check: function (e) {
+      if (e != null) {
```
markhayden: `if(!e) {` ?

sarahyablok: `e != null` checks that e is not undefined, which is what I want here. `e` is not a bool, it's an optional param containing the event.

sarahyablok:
Going to merge as admin because the codecov/patch diff is dumb (everything I added is covered, there are bits of the original exitIntent code that moved which were not).

PR #375, opened 2020-01-02
Removes more references to npm

sarahyablok:
Removes more references to npm in favor of yarn. (work started by @cbrock in #342)


sarahyablok:
@humanchimp I did. Not sure what the diff is... But I can continue to investigate... It actually might be from the dependabot PRs cause this is the first time running gulp build since merging those?

humanchimp:
you're right

PR #427, opened 2020-05-06
Ensure scroll percentage is calculated correctly

cbrock:
Prior to this commit, under certain conditions it was impossible for the
scrollPercentageToDisplay to ever trigger.

Consider the following:
- page content height of 800px
- scrollPercentageToDisplay setting of 30 (percent)
- window height of 600px

In the above scenario, the user would need to scroll 240 pixels down the
page to trigger a modal, however there is only 200px of available
scrolling. The modal will never trigger.

This commit changes the arithmetic of percentageScrolled such that in
the above scenario the modal will correctly trigger after scrolling 30%
of the available space.


src/rollup/display-conditions/scroll/register-position-watcher.js
```diff
@@ -15,14 +15,19 @@ import document from '../../dom/document';
...
+            document.body.scrollHeight,
+            document.body.offsetHeight,
+            document.documentElement.clientHeight,
+            document.documentElement.scrollHeight,
+            document.documentElement.offsetHeight
+          ),
+          offset = document.documentElement.scrollTop || document.body.scrollTop,
+          clientHeight =
+          document.documentElement.clientHeight || document.body.clientHeight,
+          percentageScrolled = (offset / (height - clientHeight)) * 100;
```
sarahyablok, line 28: Noted that `document.documentElement.clientHeight` is used in both `height` and clientHeight` is there a potential that this divisor will be zero at some point resulting in a NaN? Maybe we need to remove `clientHeight` from the max calculation for `height`? :thinking:

test/acceptance/display-conditions.spec.js
```diff
@@ -70,6 +70,40 @@ describe('when setting display conditions', function () {
...
+        scrollPercentageToDisplay: 30
+      }
+    });
+    pathfora.initializeWidgets([subscription]);
+
+    var widget = $('#' + subscription.id);
+    expect(widget.length).toBe(0);
+
+    var height = $(document.body).height();
+    window.scroll(0, height);
```
sarahyablok, line 95: Maybe we should try a partial scroll on this test too? Since the goal was to fix the math here? :thinking:

cbrock: Good idea. This test now scrolls `height / 2`.

src/rollup/display-conditions/scroll/register-position-watcher.js
cbrock: I refactored the arithmetic slightly but you're right - it is possible to end up with a `NaN` dividend in a scenario where `scrollHeight` is equal to `clientHeight`.

However, that scenario means there isn't any scrolling possible anyway, so our function should return `false`, which it does, as anything compared to `NaN` will do:

```js
  // if NaN, will always return `false`
  return percentageScrolled >= percent;
```

So, I _think_ that should be sufficient unless you'd like to be more explicit in handling `NaN` but maybe the comment is enough context?

humanchimp:
do we know what scenario `document.documentElement.xHeight` is not defined but `document.body.xHeight` is?

cbrock:
After some discussion with @humanchimp, we arrived at a solution that detects the scrolling element (either `documentElement` or `body`) based on their respective `scrollHeight` values.

Testing the case where `document.body` is the scrolling element is not straightforward but might be nice to consider in the future. The project code coverage drop appears to be because we are ignoring the hard-to-test line. Patch coverage is good.

image

PR #432, opened 2020-05-12
Fix broken docs links on lytics.github.io/pathforadocs

cbrock:
The docs link fix contained in
https://github.com/lytics/pathforajs/pull/427 didn't actually solve the
problem, but it wasn't apparent until we attempted to publish docs.

This commit fixes internal docs links to lytics.github.io/pathforadocs

### DIFF
--- a/docs/docs/customization/form.md
+++ b/docs/docs/customization/form.md
@@ -1,8 +1,8 @@
-Pathfora now allows for fully customized form fields for modules of type [form](/types/form) and [gate](/types/form). These input fields include text, texarea, select, checkbox, and radio button.
+Pathfora now allows for fully customized form fields for modules of type [form](../../types/form) and [gate](../../types/gate). These input fields include text, texarea, select, checkbox, and radio button.
 
-Try out our [form schema builder](/customization/form_builder) to easily build forms with a drag and drop interface and output a `formElements` object for your module config.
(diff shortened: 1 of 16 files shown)


docs/docs/customization/form.md
```diff
@@ -1,8 +1,8 @@
-Pathfora now allows for fully customized form fields for modules of type [form](/types/form) and [gate](/types/form). These input fields include text, texarea, select, checkbox, and radio button.
+Pathfora now allows for fully customized form fields for modules of type [form](../../types/form.md) and [gate](../../types/form.md). These input fields include text, texarea, select, checkbox, and radio button.
 
-Try out our [form schema builder](/customization/form_builder) to easily build forms with a drag and drop interface and output a `formElements` object for your module config.
+Try out our [form schema builder](./form_builder.md) to easily build forms with a drag and drop interface and output a `formElements` object for your module config.
```
sarahyablok, line 3: Is there a distinction on when we including the `.md` for some and not others? :thinking:

cbrock: My heuristic was whether the links worked correctly or not 😆. Being not very familiar with mkdocs, I didn't realize these are equivalent:

```
[form schema builder](./form_builder.md)
```
and
```
[form schema builder](../form_builder)
```

Similarly, from above in the same file:

```
[form](../types/form.md)
```
is equivalent to
```
[form](../../types/form)
```

I updated the handful of links still including `.md` to _not_ include it, adjusting the paths so the links would still work 🙃.

PR #457, opened 2020-06-12
Version 1.2.4

cbrock:
What it says on the tin.


humanchimp:
mind incrementing the version in package.json too? seems like we haven't been great about doing that

cbrock:
> mind incrementing the version in package.json too? seems like we haven't been great about doing that

Good call. Updated.

PR #538, opened 2021-02-23
feat: censorTrackingKeys config option

### LINKED ISSUE #531: Fox - FEATURE REQUEST
Fox would like to be able to use Lytics modals but not have email send Back to Lytics.  Instead they would like to send that email somewhere else. 

Additionally they would still like to capture conversions/impressions still when the user submits the modal. 

I feel like PG could also utilize this eventually

humanchimp:
Closes #531 

This PR adds a new config option `censorTrackingKeys` which takes an array containing either strings or regular expressions. When strings are used, an exact match is tested for. Otherwise, we use regular expression semantics. This can be used to censor PII in modals, for example:

```javascript
var widget = pathfora.Form({
  // .. elided
  censorTrackingKeys: [/pf-form-/, /pf-custom-form-/]
});
```

## manual testing note

this is a suffient mock for jsatg for testing this PR:

```javascript
window.jstag = {send: console.log}
```

### DIFF
--- a/docs/docs/tracking.md
+++ b/docs/docs/tracking.md
@@ -1,5 +1,38 @@
 Pathfora can send tracking events and the user data submitted to Lytics and Google Analytics. As long as your tags are all set up in the correct order, module data will be sent Lytics automatically. Below we will look at which data fields are sent in detail and their formats.
 
+## censorTrackingKeys
+
+This option exposes the capability to censor keys from payloads sent into Lytics. Any keys included as strings, or matching patterns as regular expressions that should be omitted from all payloads.
+
+<table>
+  <thead>
+    <tr>
+      <td colspan="2" align="center"><code>censorTrackingKeys</code>  array of string and/or regular expressions</td>
+    </tr>
+    <tr>
+      <th>Value</th>
+      <th>Behavior</th>
+    </tr>
+  </thead>
+
+  <tr>
+    <td>array of strings</td>
+    <td>keys which match any of the strings exactly will be censored from any payloads sent to Lytics</td>
+  </tr>
+  <tr>
+    <td>array of regular expressions</td>
+    <td>any key which matches any of the regular expression patterns will be censored from any payloads sent to Lytics</td>
+  </tr>
+</table>
+
+``` javascript
+// help ensure compliance with industry-specific PII regulations by censoring form data,
+var widget = new pathfora.Form({
+  // other configuration options elided
+  censorTrackingKeys: [/pf-form/]
+});
+```
+
 ## Lytics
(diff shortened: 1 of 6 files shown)


src/rollup/utils/censor-tracking-keys.js
```diff
@@ -0,0 +1,21 @@
...
+ * only exact matches are censored. For non-strings, if the object's test method returns true, the key will be censored.
+ *
+ * @param {object} data the data to censor
+ * @param {obejct} keysToReject an array of strings or regexps to censor the data by preparatory to sending
+ */
+export default function censorTrackingKeys (data, keysToReject) {
+  return Object.keys(data)
+    .filter(function (key) {
+      return !keysToReject.some(function (keyToReject) {
+        return typeof keyToReject === 'string'
```
ashyablok, line 12: Just checking here, so we accept mixed key types, strings and regex in the same array? I think that's a good way to go.

ashyablok (commented):
Looks good! Can we also add docs for this? Let me know if you need guidance navigating where to add them. :)

src/rollup/utils/censor-tracking-keys.js
humanchimp: correct, it detects lazily if the item is a string, so this will tolerate a mix. Here is [the spec for that case](https://github.com/lytics/pathforajs/blob/b3578284d73dc9a47abfefdb2af91a86c2104606/test/unit/utils/censor-tracking-keys.spec.js#L43)

humanchimp:
I added this to the docs:

Screen Shot 2021-02-25 at 12 20 40

Thoughts?

ashyablok:
@humanchimp Content of the docs addition looks good but I think it's better suited for this page: https://lytics.github.io/pathforadocs/tracking/

The "Generalized Configuration" page refers to an alternate object you can pass in with multiple modals to apply similar traits to all of them, so this new config option spans outside of that to any widget object. And I think the page could use a new section to accomidate this change as it is the first config option for widgets pertaining to tracking.

ashyablok (approved):

docs/docs/tracking.md
```diff
@@ -1,5 +1,38 @@
 Pathfora can send tracking events and the user data submitted to Lytics and Google Analytics. As long as your tags are all set up in the correct order, module data will be sent Lytics automatically. Below we will look at which data fields are sent in detail and their formats.
 
+## censorTrackingKeys
+
+This option exposes the capaility to censor keys from payloads sent into Lytics. Any keys included as strings, or matching patterns as regular expressions that should be omitted from all payloads.
```
krothenbaum, line 5:
```suggestion
This option exposes the capability to censor keys from payloads sent into Lytics. Any keys included as strings, or matching patterns as regular expressions that should be omitted from all payloads.
```

Capability is misspelled

krothenbaum (approved):
Pre approving once the misspelling in the docs is fixed.

PR #553, opened 2021-03-18
bug: hotfix position validation for modals.

### LINKED ISSUE #554: Exit Intent Modal for NS Media is failing
I have a client who is unable to get their experience to show. It should be showing up on exit intent: https://app.lytics.com/experiences/842115b5850c9a491696da119b9a8bc2?aid=3763&view=details

I am also seeing this error on the web page when I hover over the exit:
Screen Shot 2021-03-18 at 8 48 47 AM

ashyablok:
closes #554
Looks like the UI sets the `middle-center` position for modals. :facepalm:


docs/mkdocs.yml
```diff
@@ -1,5 +1,5 @@
 site_name: PathforaJS SDK
-nav:
+pages:
```
humanchimp, line 2: this change seems unrelated?

src/rollup/validation/validate-widget-position.js
```diff
@@ -13,7 +13,7 @@ export default function validateWidgetPosition (widget, config) {
 
   switch (config.layout) {
   case 'modal':
-    choices = [''];
+    choices = ['', 'middle-center'];
```
rallyb13, line 16: Why are you leaving empty string here?
I_believe_ modal/middle-center is a default from the initial creation of the model (the pre-selections that people can change), but they can't change it to nothing.

humanchimp:
can we change the throw to console.warn?

src/rollup/validation/validate-widget-position.js
rallyb13: Also, maybe I don't understand what's going on here, but we seem to be 
- missing cases:
`sticky-bar` (top-center, bottom-center)
`gate` (also middle-center only)
- including a case I don't recognize (button?)
- referring to the choices differently (bar is now just top-center; what's "top-absolute"?)

ashyablok: @rallyb13 The pathfora position scheme was different than the ones set up for the API. In pathfora modals don't have a position as there is only one. I have now included the missing case for  stickybar. And in Pathfora gate is considered a modal. So we should be good.

docs/mkdocs.yml
ashyablok: I noticed this yesterday when deploying docs that changing this messed up the navigation so I'm changing it back, but yes it is unrelated to this PR.

rallyb13 (approved):
As Chris said, we can fix/cleanup more as needed on our own time now that we've switched to warn. So approving even though I clearly don't have the context for the rest of this. 😬

humanchimp (approved):

PR #599, opened 2021-06-28
bug: 598 Only expire items from localStorage added by pathforajs

willphdavies:
tried to keep this backwards compatible


test/unit/persist/expiring-local-storage.spec.js
```diff
@@ -29,7 +29,8 @@ describe('`expiringLocalStorage` util', function () {
 
         expect(tuple).toEqual({
           $: 'wow',
-          '@': expiresOn.toISOString()
+          '@': expiresOn.toISOString(),
+          'PATHFORA': 'PATHFORA'
```
humanchimp, line 33: i think maybe we should do `PATHFORA: $version` instead

humanchimp (commented):
because pathfora is so broadly deployed, i feel conflicted about leaving in folks localStorage all the pathfora tokens that will now never expire, but that does seem potentially less harmful that "expiring" "tokens" not created by this lib

ashyablok (commented):
The approach here seems good! Can you run `yarn run clean` and `yarn run prod` and recommit? That will clean up the dist files.

ashyablok (approved):

humanchimp (approved):

PR #614, opened 2021-09-18
feat: allow PathforaCSS to come from outside

humanchimp:
this PR lets the user configure the URL to use for the pathfora css file by defining the global `PathforaCSS` before loading this `pathfora.min.js`


ashyablok (commented):
@humanchimp Wondering if we really want to allow the user to set CSS which would remove the base, default styles for pathfora. Usually when we have someone using custom CSS it is to override/build off of base styles - we can discuss offline/via slack, but I'm curious to know the use case for this?

ashyablok (approved):

humanchimp:
i ran the tests locally, since CI is down

PR #615, opened 2021-09-30
Add support for global total impressions and duration

### LINKED ISSUE #613: [FR] add ability to hide a modal for a definable period of time at both modal and global level
We currently have a display condition to prevent a modal from showing more than X times per session. This setting is either executable at the single modal layer (only limit this modal to one per session) or at the global layer across all modals (only allow one impression from any modal during this session).

Unfortunately, the session has proven to be problematic for a few reasons:
1. session storage resets if you open a new tab
2. session storage is not flushed on the same interval as our tag
3. sessions in general have loose definitions and understandings which leads to confusion (i thought the session was 10 minutes...I thought it was expanded only on page view interactions...etc.)

As an alternative, we would like to give control over the duration as part of the definition. So the statement becomes "hide this modal for this period of time after this number of impressions". Said rule would then adhere to both individual modal or global capabilities like session ("I want to prevent any modal from showing for 24 hours after this modal gains an impression")

The actual JSON definition would look something like this:

```
displayConditions: {
  impressions: {
    widget: {
        total: 1,
        duration: 3600 // seconds
      },
    global: {
      total: 1,
      duration: 3600 // seconds
    }
  }
}
```

In the above example, we'd need to extend the existing "total" option to support the global setting and then introduce a duration setting. if the duration was not set it would default to the existing session setting which would maintain backward compatibility.

Finally, this setting should also be added to the app UI as a manageable display condition from the Experience Wizard.

image

ashyablok:
closes #613 

Adds support for the following:
- `total` impressions key for `global` level impressions.
- `duration` impressions key for impressions using the `total` key.

Example configs:

``` javascript
// example: hide the module for 24 hours after five impressions

displayConditions: {
  impressions: {
    widget: {
      total: 5
      duration: 60 * 60  * 24,
    }
  }
}
```

```javascript
// example: hide the module for 24 hours after five impressions from **any** pathfora modules

displayConditions: {
  impressions: {
    global: {
      total: 5,
      duration: 60 * 60  * 24,
    }
  }
}
```


ashyablok:
Note: Our travis CI integration still seems to be broken D: Chris and I tried to migrate to the new Travis and they never got back to us. I will try to work on fixing this a bit more tomorrow, but for reviewers: you can run tests locally and see that they pass!

willphdavies (approved):
I don't know a huge amount about this repo, but code looks good and the tests pass! :-)

jasminebui (approved):
I set this up locally after the demo and played around with this. Looks good! I couldn't think of any weird cases that would cause this to fail. 🚀

PR #623, opened 2022-06-06
add additional prioritization rule to allow unordered in addition to ordered

markhayden:
When passing the prioritization options we currently allow for `ordered`. If any other value is passed this results in an error due to validation. This means that there is no way to override the defaults loaded along with the core Lytics tag at the tag manager layer as passing an empty string gets overridden and any other string errors. 

This adds a new valid option as `unordered` which enables proper overrides. 

image


RaeEversmann (approved):

humanchimp:
I think we need the dist in-repo?

ashyablok (approved):

PR #628, opened 2023-06-06
bug: update pathforajs to use jstag3 callbacks

RaeSmith1331:


dist/pathfora.js
```diff
@@ -1014,11 +1014,23 @@
...
     } else if (window.jstag && typeof window.jstag.getEntity === 'function') {
+
+      var version = parseInt(window.jstag.version, 10);
       // > jstag 3.0.0
-      var entity = window.jstag.getEntity();
-      if (entity.data && entity.data.user) {
-        cb(entity.data.user);
-        return;
+      if (version >= 3) {
+        window.jstag.send('entityReady', function () {
```
ashyablok, line 1021: I have never seen this format before. 🤔 could we not do:

`window.jstag.entityReady(entity => ...)`

RaeEversmann: updated

dist/pathfora.js
```diff
@@ -1014,11 +1014,29 @@
       cb(window.lio.data);
       return;
     } else if (window.jstag && typeof window.jstag.getEntity === 'function') {
-      // > jstag 3.0.0
-      var entity = window.jstag.getEntity();
-      if (entity.data && entity.data.user) {
-        cb(entity.data.user);
-        return;
+      if (window.jstag.version) {
```
RaeEversmann, line 1017: I don't think jstag2 has the version field, so we need a check here.

ashyablok (approved):

humanchimp (approved):

PR #663, opened 2025-10-20
Allow for Gate Layout and Remove Variants

ashyablok:
This PR does the following (with full backwards compatibility):
- Gates can now be made as a layout (instead of just a type). Message and Form types are supported by the gate layout.
- Removes the dependency on the variant classes for widgets with content recommendations and images.

Use `yarn run local` to run the local gulp server and test configs in the test.html file.


teijas:
When I've configured the message as layout "gate", the x button hides, but the action button is also not visible. I think it's intended that the action button should show.

image

Here's my configuration:

```
var message = new pathfora.Message({
        id: 'bar-valued-customers',
        layout: 'gate',
        msg: 'Thanks for being a valued customer, please check out our new products.',
        cancelShow: false,
        okMessage: 'View Now',
        confirmAction: {
          name: 'view now',
          callback: function () {
            window.location.pathname = '/new-products';
          },
        },
      });
```

When I switch the layout to "modal", then both buttons show:
image

teijas (approved):

markhayden (approved):

